PR #406, opened 2015-09-18
Ftr 404 update locale

jjlankinen:
Changed FI locale string as requested.


jjlankinen:
This is NOT valid pull request!

Issue #389, opened 2015-09-14
Exercise save button doesn't reliably follow changes in field values


pkukkon:
1. Have an exercise with two or more text fields. Its save button says "Hand in Exercise"
2. Write something in one field and click the save button
3. It should save, causing the button text to change to "Exercise handed in"
4. Move focus to another field and type something without leaving the field
5. The save button still reads "Exercise handed in" (bug?)
6. Clicking the save button doesn't seem to do anything (bug?)
7. Only after the field loses focus, the button changes back to "Hand in Exercise" and allows saving

All in all, it seems that if a field value has changed but the field still has focus, clicking save button seemingly does nothing. This is confusing as users essentially end up clicking the button twice.

Either there's not enough time for necessary code execution between the field losing focus and the save button receiving a click, or - more likely - there's a bug handling the state change from "Exercise handed in" back to state "Hand in Exercise." (the button text should change as soon as a field value changes rather than when the field loses focus).

PR #446, opened 2015-09-30
Fix for bug #389

pkukkon:


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-materials.js
```diff
@@ -85,7 +85,8 @@
   $(window).load(function () {
     // Workspace's material's TOC
     if ($('#workspaceMaterialsTOCWrapper').length > 0) {
-      
+
+      var contentContainer = ($('#contentWorkspaceMaterials').length > 0 ? contentContainer = $('#contentWorkspaceMaterials') : contentContainer = $('#content'));
```
ilmoeuro: These assignment expressions (the two latter "contentContainer ="'s) are redundant and confusing.

PR #529, opened 2015-10-15
Fixes #468 student records not working

### LINKED ISSUE #468: (student) records not working
Student gets code and link that is not working.
records

avukkonen:
By taking the description from data-attribute to hidden div container

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/records.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/records.js
@@ -104,10 +104,10 @@
     },
     _onItemClick: function (event) {
       var item = $(event.target).hasClass('tr-item') ? $(event.target) : $(event.target).closest('.tr-item');
-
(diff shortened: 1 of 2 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/records.js
```diff
@@ -104,10 +104,10 @@
     },
     _onItemClick: function (event) {
       var item = $(event.target).hasClass('tr-item') ? $(event.target) : $(event.target).closest('.tr-item');
-
+      
       var workspaceEntityId = $(item).attr('data-workspace-entity-id');
       var workspaceEntityName =  $(item).attr('data-workspace-entity-name');
-      var workspaceEntityDescription = $(item).attr('data-workspace-entity-description');
+      var workspaceEntityDescription = $(item).find('#description').html();
```
anttileppa: .html() vs text() ?

Issue #543, opened 2015-10-20
Guider - javascript error when object "users" is undefined


avukkonen:
Javascript runs to an error when users object is undefined

PR #544, opened 2015-10-20
Fixes the bug #543  by adding an if -block to test if "users" is undefined

avukkonen:


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/guider.js
```diff
@@ -90,7 +90,10 @@ $(document).ready(function(){
         }
         mApi().user.users.read({searchString : params, archetype : 'STUDENT', maxResults: 25 })
         .callback(function (err, users) {
-          users.userCount = users.length;         
+          
+          if(users != undefined){
```
anttileppa: It would be better to use typeof (users != undefined) also matches false, 0, empty array, etc,,,

PR #554, opened 2015-10-20
#552 Fixes the bug and adds the icons to html instead of extending them

avukkonen:


muikku/src/main/webapp/resources/theme-muikku/sass/_communicator.scss
```diff
@@ -77,8 +71,6 @@
     &.usergroup{
 
       &:before{
-         // @extend .icon-members;
-         // extend breaks icon in firefox
          content: "\e60d";
```
avukkonen: remove content

muikku/src/main/webapp/resources/theme-muikku/sass/_communicator.scss
```diff
@@ -87,12 +79,16 @@
     &.workspace{
 
       &:before{
-         // @extend .icon-books;
-         // extend breaks icon in firefox
          content: "\e61b";
```
jjlankinen: This needs to be removed as the icon is added directly to dust file as element class. Content was dirty fix for immediate problem.

Issue #812, opened 2015-10-28
Material management crashes when uploading files without extension


pkukkon:
If the content type of an uploaded file cannot be determined, material creation fails.

pkukkon:
Merged to devel.

PR #814, opened 2015-10-28
Fix for #812

pkukkon:
Changed empty content types to application/octet-stream

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-material-upload.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-material-upload.js
@@ -45,7 +45,7 @@
     _uploadFile: function (file) {
       mApi().materials.binary.create({
         title: file.name,
-        contentType: !file.contentType || file.contentType == '' ? 'application/octet-stream' : file.contentType,
+        contentType: file.contentType||'application/octet-stream',
         fileId: file.id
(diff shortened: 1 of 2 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-material-upload.js
```diff
@@ -45,7 +45,7 @@
     _uploadFile: function (file) {
       mApi().materials.binary.create({
         title: file.name,
-        contentType: file.contentType,
+        contentType: !file.contentType || file.contentType == '' ? 'application/octet-stream' : file.contentType,
```
anttileppa: file.contentType||'application/octet-stream'?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-materials-management.js
```diff
@@ -562,6 +562,9 @@
       var fileId = data._response.result.fileId;
       var fileName = data.files[0].name;
       var contentType = data.files[0].type;
+      if (!contentType || contentType == '') {
```
anttileppa: both sides of or yield to same result, so only if (!contentType) would be suffient, but using var contentType = data.files[0].type||'application/octet-stream'; would be even better.

Issue #813, opened 2015-10-28
Incorrect URL name generation for files


pkukkon:
When dealing with duplicate resource names, file names are formed incorrectly. For example, abc.jpg becomes abc.jpg-2 rather than abc-2.jpg.

pkukkon:
Merged to devel.

PR #815, opened 2015-10-28
Fix for #813

pkukkon:
Changed URL name generation so that if the name looks like a file name, the unique counter is appended before the file extension rather than the end of the name.

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
+++ b/muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
@@ -905,7 +905,7 @@ public class WorkspaceMaterialController {
             break;
           }
           // uniqueName in use, try again with the next candidate (name, name-2, name-3, etc.)
-          uniqueName = isFileName ? fileName + "-" + ++i + "." + extension : urlName + "-" + ++i; 
+          uniqueName = isFileName ? String.format("%s-%d.%s", fileName, ++i, extension) : String.format("%s-%d", urlName, ++i); 
         } else {
           // Current uniqueName is available
           break;
--- a/muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
+++ b/muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
@@ -887,6 +887,10 @@ public class WorkspaceMaterialController {
 
   public synchronized String generateUniqueUrlName(WorkspaceNode parent, WorkspaceNode targetNode, String title) {
     String urlName = generateUrlName(title);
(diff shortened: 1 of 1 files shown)


muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
```diff
@@ -887,6 +887,10 @@ public synchronized String generateUniqueUrlName(WorkspaceNode parent, String ti
 
   public synchronized String generateUniqueUrlName(WorkspaceNode parent, WorkspaceNode targetNode, String title) {
     String urlName = generateUrlName(title);
+    String fileName = StringUtils.substringBeforeLast(urlName, ".");
```
anttileppa: Wouldn't it be safer to use FilenameUtils.getBaseName?

muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
```diff
@@ -887,6 +887,10 @@ public synchronized String generateUniqueUrlName(WorkspaceNode parent, String ti
...
+    String extension = StringUtils.substringAfterLast(urlName, ".");
```
anttileppa: Wouldn't it be safer to use FilenameUtils.getExtension?

muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
```diff
@@ -901,7 +905,7 @@ public synchronized String generateUniqueUrlName(WorkspaceNode parent, Workspace
             break;
           }
           // uniqueName in use, try again with the next candidate (name, name-2, name-3, etc.)
-          uniqueName = urlName + "-" + ++i;
+          uniqueName = isFileName ? fileName + "-" + ++i + "." + extension : urlName + "-" + ++i;
```
anttileppa: Refactor to use String.format to make it more readable

muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
pkukkon: FilenameUtils.getBaseName is only suitable for dealing with file names. This code deals with titles (and occasional file names). In other words, if this method was given a string "1. luku", using FilenameUtils would turn it into name "1" and extension " luku", leading to unpredictable behaviour further down the line.

muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialController.java
pkukkon: See above.

PR #816, opened 2015-10-28
Bug 637 autosave fails if connection breaks

ilmoeuro:


muikku-core-plugins/src/main/java/fi/muikku/plugins/websocket/WebSocketPingHandler.java
```diff
@@ -0,0 +1,16 @@
+package fi.muikku.plugins.websocket;
+
+import javax.enterprise.event.Observes;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+@Singleton
```
anttileppa: Singleton?

muikku/src/main/webapp/resources/scripts/gui/websocket.js
```diff
@@ -119,6 +125,26 @@
       return null;
     },
     
+    _startPinging: function () {
+      this._pingHandle = setInterval($.proxy(function() {
```
anttileppa: Usage of setInterval instead of setTimeout could potentially lead to concurrency problems

muikku-core-plugins/src/main/java/fi/muikku/plugins/websocket/WebSocketPingHandler.java
ilmoeuro: What else? There could be no request or session active. There's no state, so no requirement for multiple instances either.

PR #871, opened 2015-11-04
Ftr 531 non relevant users

### LINKED ISSUE #531: User search finds non-relevant users
When searching users, all the users from school data source show up, even if they are not supposed to be seen by the searching user.

anttiviljakainen:
Fixes #531


anttiviljakainen:
Still needs fixing on 
- empty list of published workspaces
- test workspace is probably not published

PR #929, opened 2015-11-16
Ftr 928 ws ip filter

### LINKED ISSUE #928: Remove websocket ticket ip filter
Users with poor mobile connects seem to receive new ip addresses quite often which causes lot of troubles for with web socket connections

anttileppa:
Fixes #928


anttileppa:
Master...

PR #1030, opened 2015-12-04
Ftr 990 evaluation answer element size

### LINKED ISSUE #990: Student assignment answers element too small
As a teacher in evaluation I would like to see students assignment answer element height adjusted based on its content so it would be easier to read and evaluate students answer.

jjlankinen:
FIxes #990


muikku-evaluation-plugin/src/main/resources/META-INF/resources/scripts/gui/evaluation.js
```diff
@@ -284,6 +283,19 @@
       var content = $(assignment).find('.evaluation-assignment-content');
       var state = content.attr('data-open-state');
       content.attr('data-open-state', state == 'closed' ? 'open' : 'closed');
+      this._adjustTextareaHeight(content);
+    },
+    
+    _adjustTextareaHeight: function(container) {
+      var textareas = $(container).find('.muikku-memo-field');
+      $(textareas).each( function(index,textarea) {        
+        $(textarea).css({
```
anttileppa: Why reduce the size to just 1 when next clause will override it anyway?

PR #1138, opened 2015-12-17
Oneditannouncement click broken, editable content loading

### LINKED ISSUE #1132: _onEditAnnouncementClick broken
In announcer.js, event.target is null in _onEditAnnouncementClick

avukkonen:
Fixes #1132


muikku-core-plugins/src/main/resources/META-INF/resources/dust/announcer/announcer_item.dust
```diff
@@ -1,5 +1,5 @@
 
-<div class="mf-item an-announcement">
+<div class="mf-item an-announcement" id="{id}">
```
anttileppa: it's bad practice to use database row id as html element id, maybe you should use data-announcement-id instead?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/announcer.js
```diff
@@ -73,7 +73,7 @@
       var formFunctions = function() {        
         
         // Date and timepickers for start and end time/date
-        mApi()
+        mApi({async: false})
```
anttileppa: You should async: false only if it's absolutely necessary

PR #1163, opened 2015-12-18
Ftr 1092 workspace material access

### LINKED ISSUE #1092: Public (login not required) materials view in workspace
Required for help workspace

anttiviljakainen:
Fixes #1092


muikku-core-plugins/src/main/java/fi/muikku/plugins/workspace/WorkspaceMaterialsBackingBean.java
```diff
@@ -54,7 +59,7 @@
   private SystemSettingsController systemSettingsController;
 
   @RequestAction
-  public String init() {
+  public String init() throws UnsupportedEncodingException {
```
anttileppa: This can cause an EJBRemovedException, remove throws and replace with return NavigationRules.INTERNAL_ERROR && log

PR #1170, opened 2015-12-21
Calendar timezone

### LINKED ISSUE #191: Wrong times in calendar
When I add a new calendar entry or move it in Google Calendar, the time of the event is one or two hours off.

avukkonen:
Fixes #191
Fixes Datepicker bug in chrome

### DIFF
--- a/muikku-calendar/src/main/resources/META-INF/resources/scripts/gui/calendar.js
+++ b/muikku-calendar/src/main/resources/META-INF/resources/scripts/gui/calendar.js
@@ -61,17 +61,15 @@
            delete values.allDay;           
            values.allDay = 'false';           
          }         
+         var timezone = jstz.determine().name();
+         
          values.recurrence = recurrence;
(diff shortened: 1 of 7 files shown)


muikku-calendar/src/main/resources/META-INF/resources/scripts/gui/calendar.js
```diff
@@ -65,8 +65,8 @@
          values.start = eventStart;
          values.end = eventStop;
          values.status = 'CONFIRMED';
-         values.startTimeZone = 'GMT+3';
-         values.endTimeZone = 'GMT+3';        
+         values.startTimeZone = 'GMT+2';
+         values.endTimeZone = 'GMT+2';
```
anttiviljakainen: These should most likely be "localized" too as this change i believe is based on daylight savings. More concretely, this setting should come from the users' current timezone.

PR #1395, opened 2016-02-02
** Added support changing course participation types in Pyramus when

### LINKED ISSUE #1388: Change Pyramus course student state to evaluated on evaluation
As a teacher i would like that Pyramus course student state would be automatically changed to evaluated when i evaluate student with passing grade

anttileppa:
Fixes #1388


muikku-school-data-pyramus/src/main/java/fi/muikku/plugins/schooldatapyramus/CourseParticipationTypeEvaluationMapper.java
```diff
@@ -0,0 +1,111 @@
...
+    if (grade.getPassingGrade() == null) {
+      logger.log(Level.SEVERE, "passingGrade property is null");
+      return null;
+    }
+    
+    Map<String, Long> defaults = config.getDefaults();
+    if (defaults == null) {
+      logger.log(Level.SEVERE, "Cannot resolve grade participantion type because specific grade mapping could not be found and defaults are not defined");
+    }
+
```
pkukkon: Should return if defaults is null

PR #1407, opened 2016-02-04
** Fixed issue #1399

### LINKED ISSUE #1399: Locales when browser language is "other"
When browser language is something else than Finnish or English, the environment does not recognize it and becomes confusing. 

"Other" should always be Finnish or English.

Stakeholder: Jaana Matkaselkä
osTikcket: 359717

anttileppa:
Fixes #1399

### DIFF
--- a/muikku-core/src/main/java/fi/muikku/session/local/LocalSessionControllerImpl.java
+++ b/muikku-core/src/main/java/fi/muikku/session/local/LocalSessionControllerImpl.java
@@ -12,6 +12,7 @@ import javax.enterprise.context.SessionScoped;
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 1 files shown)


muikku-core/src/main/java/fi/muikku/session/local/LocalSessionControllerImpl.java
```diff
@@ -31,6 +32,7 @@
 public class LocalSessionControllerImpl extends AbstractSessionController implements Serializable, LocalSessionController {
   
   private static final long serialVersionUID = 4947154641883149837L;
+  private static final String[] SUPPORTED_LOCALES = {"fi", "en"};
```
hkurhinen: Maybe in future this should be made configurable but this is ok for now.

PR #1563, opened 2016-02-26
** Fixes #1562

### LINKED ISSUE #1562: "No files" shown in file field if you add files and submit exercise

ilmoeuro:
** Fixes #1562


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-file-field-component.js
```diff
@@ -312,11 +312,11 @@
       this._readonly = readonly;
       if (readonly) {
         $(this._uploader).attr("disabled", "disabled");
-        if (this.element.data('file-count') > 0) {
```
anttileppa: Does this work when removing files? As far as i remember removing files does not actually remove element but marks it as removed instead (so it can be restored)

PR #1573, opened 2016-02-29
Ftr 1522 responsive workspace frontpage

### LINKED ISSUE #1522: Responsive workspace frontpage

jjlankinen:
Fixes #1522


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace.js
```diff
@@ -73,25 +65,14 @@
...
-      refreshNavigationWrapperPosition();
-      $(window).resize(function(){
-        refreshNavigationWrapperPosition();
-      });
-    }
-    
-    if ($('.wi-workspace-dock-navi-button-cancel-evaluation'.length > 0)) {
-      $('.wi-workspace-dock-navi-button-cancel-evaluation').hide();
+
+    if ($('.workspace-dock-navi-button-cancel-evaluation'.length > 0)) {
```
anttileppa: 1) This evaluates always as true 2) What is the purpose of this if? ($('.workspace-dock-navi-button-cancel-evaluation') is a set, so if there are no elements, hide just does not do anything)

PR #1911, opened 2016-04-26
Bug 1893 coursepicker paging

### LINKED ISSUE #1893: Course picker paging is replacing course list instead of adding more to the list

avukkonen:
Fixes #1893


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/coursepicker.js
```diff
@@ -160,9 +165,10 @@
       window.location = CONTEXTPATH + '/workspace/' + workspaceUrl;
     }, 
     
-    _loadMore: function () {
+    _loadMore: function () {        
       this._firstResult += this._maxResults;
-      this._loadWorkspaces();
+      hasPage = true;
+      this._loadWorkspaces(hasPage);
```
pkukkon: Variable hasPage isn't really needed here; this._loadWorkspaces(true); would be sufficient. Alternatively, local variables should be declared as such (var hasPage = true)

PR #2006, opened 2016-05-12
Ftr 1332 discussions quoting a message

### LINKED ISSUE #1332: Discussions - quoting a message
##### Stakeholders

 - Tommi Issakainen
 - Tuija Salminen

##### Details

 - When replying to the original message (OP), show the reply in top level
 - When replying to some other message, show the reply indented (as a sub-discussion)
 - Opens an editor containing the quoted text, so that the user can edit the text

avukkonen:
Closes #1332


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/discussion.js
```diff
@@ -435,10 +437,11 @@ $(document).ready(function() {
 
     _replyMessage : function(event) {
       var element = $(event.target);
+      var quote = element.parent().hasClass("di-message-quote-link") ? true : false;
```
avukkonen: use closest()

PR #2059, opened 2016-05-24
Ftr 2056 jsf roles

### LINKED ISSUE #2056: Remove isRole methods from jsf
Jsf code should not contain SessionBackingBean.teacherLoggedIn, etc calls at all

anttileppa:
Fixes #2056


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/WorkspaceHelpPageBackingBean.java
```diff
@@ -60,6 +65,11 @@ public String init() {
     if (workspaceEntity == null) {
       return NavigationRules.NOT_FOUND;
     }
+    
+    if (!sessionController.hasWorkspacePermission(MuikkuPermissions.MANAGE_WORKSPACE_HELP, workspaceEntity)) {
```
pkukkon: This backing bean is responsible of showing workspace help instead of managing it, so this permission check is invalid.

PR #2126, opened 2016-06-01
Correct answers in evaluation

### LINKED ISSUE #2098: Show the number of correct answer *fields* in evaluation view

pkukkon:
- resolves #2098


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
```diff
@@ -127,7 +126,7 @@
           }, this));
         }
         else {
-          $(pageElement).append(parsed);
+          $(pageElement).muikkuMaterialPage('content', parsed);
```
pkukkon: Append parsed children

PR #2248, opened 2016-07-01
Ftr 2189 curriculum filter

### LINKED ISSUE #2189: Curriculum filter
As a student I would be able to filter workspaces by curriculum in course picker. Default should be my default curriculum

anttiviljakainen:
Fixes #2189


anttiviljakainen:
This update needs guidance text or it will confuse students as the default filter kicks in in the course picker.

PR #2268, opened 2016-08-03
Blocks the search when the announcer is already chosen.

### LINKED ISSUE #2063: Announcer - same recipient multiple times
In the announcer, there is no check if the recipient is already in the list.

avukkonen:
fixes #2063


anttiviljakainen:
Check the most obvious hound stuff next time.

PR #2345, opened 2016-08-24
Fix for UserSchoolDataBridge not handling null staff members properly

### LINKED ISSUE #2344: UserSchoolDataBridge fails to handle archived staff members
For archived staff members, UserSchoolDataBridge tries to create an entity from a null User.

pkukkon:
- fixes #2344


anttiviljakainen:
Looks ok, merge after tests.

PR #2350, opened 2016-08-24
Plugin setting for filtering groups shown in workspace permissions view

### LINKED ISSUE #2348: Setting for limiting groups shown in workspace permissions view
Workspace permissions view currently lists practically all study programs and user groups. It should be possible to configure the view so that only "relevant" study programs and user groups would be listed.

#### OSTicket
http://helpdesk.muikkuverkko.fi/scp/tickets.php?id=1073

#### SH
Nina Kurki

pkukkon:
- resolves #2348


anttiviljakainen:
Things are good here, merge after tests.

PR #2358, opened 2016-08-29
Changed sorter correct answer checking to cope with identical words

### LINKED ISSUE #2322: Sorter field - identical words should be equivalent
##### Details

 - When there's multiple identical words in sorter field, they should be equivalent when grading the exercise

pkukkon:
- resolves #2322


anttiviljakainen:
Looks ok, merge after tests.

PR #2360, opened 2016-08-30
Connection field duplicate counterpart support

### LINKED ISSUE #2321: Connection field -  identical fields should be equivalent
##### Details

 - When there's multiple identical items in connection field, they should be equivalent when grading the exercise

pkukkon:
- resolves #2321


anttiviljakainen:
Looks ok, merge after tests.

PR #2415, opened 2016-09-14
Fix for workspace copying "breaking" source workspace

### LINKED ISSUE #2368: Course copy breaks original course
After course copy, the original course's settings cannot be updated/saved. 

Course settings of an copy can be updated/saved properly.

It does not matter if materials are copied or linked.

pkukkon:
- fixes #2368


pkukkon:
Some functionality still needed for copying workspaces.

PR #2438, opened 2016-09-21
 - Added duplicate recipient cleaning to message sending

### LINKED ISSUE #2375: Duplicate recipients in communicator messages
When sending communicator messages, the list of recipients is constructed without checking for duplicates. This can lead to the same message being delivered to the same recipient multiple times.

anttiviljakainen:
Fixes #2375

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/CommunicatorController.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/CommunicatorController.java
@@ -344,7 +344,7 @@ public class CommunicatorController {
    * @param userEntities
    * @return
    */
-  public List<UserEntity> cleanDuplicateRecipients(List<UserEntity> userEntities) {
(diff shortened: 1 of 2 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/CommunicatorController.java
```diff
@@ -337,4 +338,27 @@ public void delete(CommunicatorUserLabel communicatorUserLabel) {
     communicatorUserLabelDAO.delete(communicatorUserLabel);
   }
 
+  /**
+   * Cleans list of UserEntities so that there are no duplicates present. Returns the original list.
+   * 
+   * @param userEntities
+   * @return
+   */
+  public List<UserEntity> cleanDuplicateRecipients(List<UserEntity> userEntities) {
```
pkukkon: Should return void to better indicate that the given list is the one that the method modifies.

PR #2505, opened 2016-10-06
Bug 2440 recipients in sent messages

### LINKED ISSUE #2440: Recipients in sent messages
As a user, I would like to see the recipients in my sent messages. Now I only see my own name.

### LINKED ISSUE #2498: Coursepicker load animation doesn't show
There doesn't seem to be css in place for coursepicker loading animation

avukkonen:
Fixes #2498 
Fixes #2440


pkukkon:
Cannot be merged as-is. If sending a message to a large user group, the sent message lists each individual recipient, and also makes a separate call to the school data source for each.

pkukkon:
Cannot be merged as-is. When replying to a large user group message, the entire list of recipients is loaded to message editor. With enough recipients, the message user interface grows so big that it won't fit on screen and won't scroll either.

pkukkon:
Then again, looks like the message editor always loads all recipients, so large group messages are already a problem... Needs an overhaul but can probably be merged after conflicts have been resolved.

PR #2526, opened 2016-10-11
Ftr 2432 remove flag from your flag list

### LINKED ISSUE #2221: Deleting flags
As a a teacher I would be able to either remove flags manually or that they would just disappear if none of the students is flagged by using that particular flag.

##### Details

 - What if the flag is shared and others have added students to it?
  - The flag is just removed, and it disappears from everyone

### LINKED ISSUE #2432: Remove a flag from your flags list
##### Details

 - A non-owner can remove the flag from himself, it doesn't disappear from others (and the owner)

ilmoeuro:
** Fixes #2432 
** Fixes #2221


muikku-rest/src/main/java/fi/otavanopisto/muikku/rest/user/UserRESTService.java
```diff
@@ -701,6 +702,54 @@ public Response updateFlag(@PathParam ("ID") Long id, fi.otavanopisto.muikku.res
...
+            loggedIdentifier);
+    
+    if (loggedUserIdentifier == null) {
+      return Response
+                .status(Status.BAD_REQUEST)
+                .entity("No user school data identifier for logged user")
+                .build();
+    }
+    
+    if (Objects.equals(ownerIdentifier.getIdentifier(), loggedUserIdentifier.getIdentifier())) {
```
anttiviljakainen: In theory identifiers could match, but come from different data sources.

anttiviljakainen: On the other hand, in scripts owner is checked against userentity while here it's against identifier. Is this wanted functionality?

PR #2718, opened 2016-12-08
Restore latest evaluation request (if any) when workspace evaluation is removed

### LINKED ISSUE #2716: Restore latest evaluation request when evaluation is removed
If a teacher evaluates a workspace based on an evaluation request, the request is marked as handled and thus disappears from the request list. If the evaluation was made in error and removed, the request remains handled. Request should be marked not handled when the evaluation is removed.

pkukkon:
- resolves #2716

### DIFF
--- a/muikku-core/src/main/java/fi/otavanopisto/muikku/schooldata/GradingController.java
+++ b/muikku-core/src/main/java/fi/otavanopisto/muikku/schooldata/GradingController.java
@@ -180,15 +180,17 @@ public class GradingController {
     });
     // Update should cause the school data source to treat the request as active again (i.e. not handled)
     WorkspaceAssessmentRequest latestRequest = requests.isEmpty() ? null : requests.get(0);
-    updateWorkspaceAssessmentRequest(
-        latestRequest.getSchoolDataSource(),
-        latestRequest.getIdentifier(),
-        latestRequest.getWorkspaceUserIdentifier(),
-        latestRequest.getWorkspaceUserSchoolDataSource(),
-        workspaceIdentifier.getIdentifier(),
-        studentIdentifier.getIdentifier(),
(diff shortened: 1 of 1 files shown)


muikku-core/src/main/java/fi/otavanopisto/muikku/schooldata/GradingController.java
```diff
@@ -170,6 +171,26 @@ public WorkspaceAssessment updateWorkspaceAssessment(SchoolDataIdentifier worksp
...
     gradingSchoolDataController.deleteWorkspaceAssessment(workspaceIdentifier, studentIdentifier, workspaceAssesmentIdentifier);
+    // #2716: When workspace assessment is removed, restore latest workspace assessment request
+    List<WorkspaceAssessmentRequest> requests = listWorkspaceAssessmentRequests(workspaceIdentifier.getDataSource(), workspaceIdentifier.getIdentifier(), studentIdentifier.getIdentifier());
+    requests.sort(new Comparator<WorkspaceAssessmentRequest>() {
+      public int compare(WorkspaceAssessmentRequest o1, WorkspaceAssessmentRequest o2) {
+        return o2.getDate().compareTo(o1.getDate()); // latest request first
+      }
+    });
+    // Update should cause the school data source to treat the request as active again (i.e. not handled)
+    WorkspaceAssessmentRequest latestRequest = requests.isEmpty() ? null : requests.get(0);
```
anttiviljakainen: Can requests be null? CollectionUtils might have safer isempty check.

pkukkon: By good coding practice, list methods should always return an empty list rather than null but it's admittedly better be safe than NPE.

PR #2725, opened 2016-12-08
Ftr 2668 journal paging

### LINKED ISSUE #2668: Poor performance of workspace journal for faculty
By design, faculty sees all journal entries of all students when accessing workspace journal. If there are many, the performance is appalling.

#### Details

 - Solve this problem with paging

anttiviljakainen:
Closes #2668 

Journal is sort of an hybrid of jsf and js, even though after this it's more about js. At some point the excess jsf should be removed for cleaner code.


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-journal.js
```diff
@@ -1,4 +1,76 @@
...
+          if (err) {
+            $('.notification-queue').notificationQueue('notification', 'error', err);
+          } else {
+            var template = this.options.canListAllEntries ? 'journal/journalentries_teacher.dust' : 'journal/journalentries_my.dust';
+            
+            renderDustTemplate(template, journalEntries, function(text) {
+              $("#journalEntries").append(text);
+            });
+            
+            if (journalEntries && (journalEntries.length < this.options.pageSize))
```
pkukkon: Hardly an issue but instead of disabling the button, could it simply be hidden if there are no more pages to load?

muikku/src/main/webapp/resources/theme-muikku/sass/flex/workspace-journal.scss
```diff
@@ -150,15 +152,17 @@
   
   .workspace-journal-poster-image {
     @incude border(2px,#fff,solid);
```
pkukkon: Shouldn't this be @include?

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/workspace/WorkspacePluginMessages_fi.properties
```diff
@@ -39,10 +39,8 @@ plugin.workspace.discussions.pageTitle = Keskustelu
 plugin.workspace.journal.pageTitle = Oppimisp�iv�kirja
 plugin.workspace.journal.studentHasNoEntries = Opiskelijalla ei ole yht��n p�iv�kirjamerkint��
 plugin.workspace.journal.noEntries = Sinulla ei ole viel� yht��n p�iv�kirjamerkint��.
+plugin.workspace.journal.loadMoreEntries = Lataa lis�� p�iv�kirjamerkint�j�
```
pkukkon: Kind of a long text for a button. Maybe just "Lataa lisää" as it should be a given that they will be journal entries?

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
```diff
@@ -2180,14 +2180,19 @@ else if (workspaceStudentId != null) {
           userEntity = userEntityFromWorkspaceUser;
         }
       }
-      entries = workspaceJournalController.listEntriesByWorkspaceEntityAndUserEntity(workspaceEntity, userEntity);
+      entries = workspaceJournalController.listEntriesByWorkspaceEntityAndUserEntity(workspaceEntity, userEntity, firstResult, maxResults);
     }
     
     for (WorkspaceJournalEntry entry : entries) {
+      UserEntity entryUserEntity = userEntityController.findUserEntityById(entry.getUserEntityId());
+      User user = userController.findUserByUserEntityDefaults(entryUserEntity);
```
pkukkon: Will most likely cause a call to Pyramus but I assume moderate page size means this won't cause performance issues?

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
```diff
@@ -2122,9 +2122,9 @@ public Response deleteWorkspaceStudent(@PathParam("WORKSPACEENTITYID") Long work
   public Response listJournalEntries(
       @PathParam("WORKSPACEID") Long workspaceEntityId,
       @QueryParam("userEntityId") Long userEntityId,
-      @QueryParam("workspaceStudentId") String workspaceStudentId
-  ) {
-    
+      @QueryParam("workspaceStudentId") String workspaceStudentId,
+      @QueryParam("firstResult") @DefaultValue ("0") Integer firstResult, 
+      @QueryParam("maxResults") @DefaultValue ("25") Integer maxResults) {
```
pkukkon: Purely cosmetic but this method defaults to 25 whereas the backing bean defaults to 20. Would improve Feng Shui if both sides shared the same value, whichever it is :)

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/WorkspaceJournalBackingBean.java
```diff
@@ -209,6 +215,16 @@ public Long getStudentId() {
     return studentId;
   }
   
+  public Long getPage() {
+    if (page == null)
```
pkukkon: Could also be just `return page == null ? 1l : page;` but it's admittedly a tomato/potato kind of thing.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
anttiviljakainen: That's the assumption, at least it's less than for all which it did before (albeit in the jsf/backingbean side)

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-journal.js
anttiviljakainen: I think this is the same way it works in communicator.

PR #3124, opened 2017-05-03
Only show workspace teachers on workspace front page

### LINKED ISSUE #3111: Only show workspace teachers (not other workspace staff members) on workspace front page

pkukkon:
- resolves #3111

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
@@ -1079,11 +1079,9 @@ public class WorkspaceRESTService extends PluginRESTService {
         
         // #3111: Workspace staff members should be limited to teachers only. A better implementation would support specified workspace roles
         
-        if (workspaceEntity != null) {
-          WorkspaceUserEntity workspaceUserEntity = workspaceUserEntityController.findWorkspaceUserByWorkspaceEntityAndUserEntity(workspaceEntity, userEntity);
-          if (workspaceUserEntity == null || workspaceUserEntity.getWorkspaceUserRole().getArchetype() != WorkspaceRoleArchetype.TEACHER) {
-            continue;
-          }
+        WorkspaceUserEntity workspaceUserEntity = workspaceUserEntityController.findWorkspaceUserByWorkspaceEntityAndUserEntity(workspaceEntity, userEntity);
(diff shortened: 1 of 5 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
```diff
@@ -1075,6 +1076,16 @@ public Response listWorkspaceStaffMembers(@PathParam("ID") Long workspaceEntityI
       
       if (user != null) {
         UserEntity userEntity = userEntityController.findUserEntityByUser(user);
+        
+        // #3111: Workspace staff members should be limited to teachers only. A better implementation would support specified workspace roles
+        
+        if (workspaceEntity != null) {
```
anttiviljakainen: Unnecessary check (checked above), also makes the control flow a bit more complex to understand (should the workspaceStaffMembers.add be inside this if clause?).

muikku-rest/src/main/java/fi/otavanopisto/muikku/rest/user/UserRESTService.java
```diff
@@ -1337,6 +1344,15 @@ public Response searchStaffMembers(
             }
           }
           
+          // #3111: Workspace staff members should be limited to teachers only. A better implementation would support specified workspace roles
+          
+          if (workspaceEntity != null) {
```
pkukkon: This method is used to search for staff members in general or staff members of a single workspace. In the latter case, each user's workspace role has to be manually checked as the user's role in each workspace they belong to is not stored in the Elastic index.

We're essentially trying to figure out whether the staff member belonging to a workspace is actually a teacher in that workspace. If so, they are included in the final result set. If they are, for example, just workspace tutors, they're not included.

Granted, the aforementioned logic should not even reside within a generic search function like this. Still, this was the simplest way to only show workspace teachers on the front page of a workspace.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/rest/WorkspaceRESTService.java
pkukkon: Yup, the null check is unnecessary. Other than that, this has the same logic as above. This method is used in the view that shows all the teachers and students of a workspace.

Later on, we should probably refactor the code so that all views would use the same method to retrieve the staff members of a workspace.

muikku-rest/src/main/java/fi/otavanopisto/muikku/rest/user/UserRESTService.java
anttiviljakainen: Fiiiiine.

PR #3137, opened 2017-05-09
Send communicator messages to teachers from workspace front page

### LINKED ISSUE #3112: Send communicator messages to teachers from workspace front page
#### Details

 - There's a button for each teacher in workspace teachers list
 - Clicking the button shows a message dialog in the bottom of the page similar to the one in communicator, but with no recipient shown

pkukkon:
- resolves #3112
- also shows teachers' future vacation period


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-frontpage.js
```diff
@@ -2,9 +2,97 @@
...
+            var bn = b.lastName + ' ' + b.firstName;
+            return an < bn ? -1 : an == bn ? 0 : 1;
+          });
+          for (var i = 0; i < staffMembers.length; i++) {
+            var props = staffMembers[i].properties;
+            if (props['profile-vacation-start'] && props['profile-vacation-end']) {
+              var begin = moment(props['profile-vacation-start']).startOf('day');
+              var end = moment(props['profile-vacation-end']).startOf('day');
+              var now = moment().startOf('day');
+              if (now.isSame(begin) || now.isBefore(begin) || now.isBetween(begin, end)) {
```
ilmoeuro: This doesn't show anything on the last day of vacation. `now.isSame(end) || now.isBefore(end)` should do the trick.

PR #3157, opened 2017-05-16
Word counter for memo and rich memo fields

### LINKED ISSUE #3120: Word count for memo fields
As a student I would like see the number of words while I'm answering the memo field assignments.

#### Details

 - The word count is visible in all memo fields and rich text fields

pkukkon:
- resolves #3120

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
@@ -530,6 +530,29 @@
           memoFieldElement.addClass('ckeditor-field');
         }
         $(object).replaceWith(memoFieldElement);
+        
+        // #3120 memo word counter (non-richedit)
+        if (data.meta.richedit != true) {
+          var wordCountContainer = $('<div class="word-count-container">')
+            .append($('<span class="word-count-title">').text(getLocaleText('plugin.workspace.memoField.wordCount')))
+            .append('<span class="word-count">');
+          memoFieldElement.after(wordCountContainer);
+          var countMethod = function() {
+            var text = memoFieldElement.val().trim();
+            $(wordCountContainer).find('.word-count').text(text == '' ? 0 : text.split(/\s+/).length);
+          };
+          var timer = undefined;
+          memoFieldElement.on('focus', $.proxy(function() {
+            timer = setInterval($.proxy(function() {
+              countMethod();
+            }, this), 1000);
+          }, this));
+          memoFieldElement.on('blur', $.proxy(function() {
+            countMethod();
+            clearInterval(timer);
+          }, this));
+          countMethod();
+        }
       }
     }
   });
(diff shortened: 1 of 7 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
```diff
@@ -530,6 +530,29 @@
...
+        
+        // #3120 memo word counter (non-richedit)
+        if (data.meta.richedit != true) {
+          var wordCountContainer = $('<div class="word-count-container">')
+            .append($('<span class="word-count-title">').text(getLocaleText('plugin.workspace.memoField.wordCount')))
+            .append('<span class="word-count">');
+          memoFieldElement.after(wordCountContainer);
+          var countMethod = function() {
+            var text = memoFieldElement.val().trim();
+            $(wordCountContainer).find('.word-count').text(text == '' ? 0 : text.split(/\s+/).length);
```
ilmoeuro: Use `===` instead of `==`. For example, `0 == ''` is `true`.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
```diff
@@ -530,6 +530,29 @@
...
+          };
+          var timer = undefined;
+          memoFieldElement.on('focus', $.proxy(function() {
+            timer = setInterval($.proxy(function() {
```
ilmoeuro: Is it ever possible for there to be two `focus` events with no `blur` event in between? For cases like that, I'd check and clear the timer first.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
```diff
@@ -530,6 +530,29 @@
...
+              countMethod();
+            }, this), 1000);
+          }, this));
+          memoFieldElement.on('blur', $.proxy(function() {
+            countMethod();
+            clearInterval(timer);
```
ilmoeuro: If the element is initially on focus, blur will trigger `clearInterval(undefined)`. I'd initialize it with null and null check it before clearing.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-rich-memo-field-component.js
```diff
@@ -57,15 +63,23 @@
         
       }
     },
+    
+    // #3120 memo word counter (functionality)
+    _countWords: function() {
+      var text = $(this._editor.getData()).text();
+      $(this._wordCountContainer).find('.word-count').text(text == '' ? 0 : text.split(/\s+/).length);
```
ilmoeuro: Use `===` instead of `==`.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
pkukkon: Fiiiine,.. Although the memo value is always a string, so the comparison is valid as-is. While 0 == '' is certaily true, e.g. '0' == '' is still false.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/muikku-material-loader.js
pkukkon: Can't see how this could ever happen, considering all the related fields have both focus and blur set. No biggie to check, though.

Issue #3028, opened 2017-03-14
Connect field's correct answers indicators don't update properly


jjlankinen:
When user checks connect fields answers and gets indicators in case or having wrong answers (those numbers inside a circle) and then fixes his answers and rechecks the exercise those indicators won't get removed or reset.

#### Details

 - This happens when a teacher adds a connect field, publishes the material, then adds new terms to the material and re-publishes.
 - The new fields don't get unpacked to the database, because each material field is processed only once.
 - Modifying material fields after publishing is problematic anyway, this should be discussed with product owner.

#### OSTicket
http://helpdesk.muikkuverkko.fi/scp/tickets.php?id=1570

#### SH
Nina Kurki

onzag:
After giving it a check it was found out that it was a serverside problem that was caused when an already updated resource was republished as the new values would not get registered.

For instance let's say you have created and published a connect field that has 3 values and then want to modify it to have 10; the remaining new options will fail and will be shuffled because the server cannot quite register those new responses.

It's simple to duplicate by publishing modifying and republishing a resource and can be tested by using the query in the database:

`select * from queryconnectfieldoption qcfo left join queryconnectfieldterm qcft on qcft.id = qcfo.id where qcfo.field_id = ?;` where ? is your query connect field id.

One solution we could apply is to block the already published connect fields from editing.

onzag:
Fixed at https://github.com/otavanopisto/muikku/pull/3164 please check comments.

ninakurki:
I think there is again two problems with the connect field exercise:

1. Adding more terms to published connect field exercise > happens rarely and I think we can't block exercises from editing so we can leave this as it is.

2. And this is what I have reported here: Student answer the connect field exercise and gets indicators if the answers are right or wrong. Student corrects his answers and check again. At this point indicators won't get removed or reset. So the connect field exercise is as it is and none is editing it between students answers.

![yhdistely](https://cloud.githubusercontent.com/assets/26649131/26349152/9d8550d4-3fb7-11e7-8eef-a90c32ecaa66.png)

PR #3164, opened 2017-05-17
Added hide indicators methods and hide indicators during refresh

onzag:
Took me a while to figure out what was going on but at the end I devised a solution.

This fixes the actual #3028 mutta there's a situation there that the task was initially mistaken for another and there's actually a bug when updating and publishing resources, this is not part of the scope of that problem but it was researched and a spin off task must be created that contains such data; this is a backend issue that has to be discussed in order to find a solution.


anttiviljakainen:
Code-wise I don't see any problems here. Pasi knows the fields better I believe so maybe he can take a look also.

PR #3181, opened 2017-05-23
Fixed course discussions button not working

### LINKED ISSUE #3173: Course Discussion: Load more messages -button does not work
Heli Viitanen wrote via helpdesk:

> https://otavanopisto.muikkuverkko.fi/workspace/ena6-tiede,-talous-ja-tekniikk_1/discussions#a131/591
> 
> Kyseessä ENA6-kurssin keskustelualue ja oma viestiketjuni, jossa on 38 kommenttia. En pääse lukemaan kuin vanhimmat. Mitään ei tapahdu, kun klikkaan Lataa lisää viestejä. Jos klikkaan sitä uudestaan, koko Lataa lisää viestejä -palkki häviää. Näytössä käväisee Odotetaan palvelinta.
> 
> Tämä on toiminut aiemmin muttei nyt ainakaan muutamaan päivään. En siis pääse nyt lukemaan opiskelijoiden tehtävien tätä osaa. 
> 
> Heli

Load more messages -button does not work. This problem is in the course discussion.

onzag:
This issue was actually all over the discussions but only appeared under a very specific circumstance that was that the parent reply was just cutting the children replies because the request didn't load more. eg.

```
[1]
[2]
  [3]
[4]
[5]
---------
  [6]
```

Because they were loaded in a separated manner, the 6 answer wouldn't relate to the 5th and so it wouldn't be able to create itself, causing it to crash.

Fixed by adding an accumulative variable that would remember the previous values as well as adding a failsafe mechanism for even when the parent thread does not exist.

Fixes #3173

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/discussion.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/discussion.js
@@ -48,7 +48,9 @@
     },
 
     loadThreadRepliesDetails: function(replies, callback) {
-      var replyCreatedMap = {};
+      this.replyCreatedMap = this.replyCreatedMap || {};
+      var replyCreatedMap = this.replyCreatedMap;
+      
       $.each(replies, function (index, reply) {
         replyCreatedMap[reply.id] = reply;
       });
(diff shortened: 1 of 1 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/discussion.js
```diff
@@ -94,6 +97,8 @@
                   replyParentCreatorFullName = generateFullName(usr);
                 }
               });
+              
+              replyParentTime = formatDate(moment(replyCreatedMap[reply.parentReplyId].created).toDate()) + ' ' + formatTime(moment(replyCreatedMap[reply.parentReplyId].created).toDate());
```
onzag: It is...

PR #3196, opened 2017-05-29
Further fine-tuning of workspace student activity

### LINKED ISSUE #3190: Workspaces are incorrectly associated with UserEntity
Several places list workspaces by UserEntity, which essentially means workspaces across students' every study programme. Functionality should be limited to the student's active study programme instead.

pkukkon:
- fixes #3190

Changed all WorkspaceUserEntityController methods that accept a UserEntity to use the default identifier of that user. For example, asking for all workspaces of a user only returns those that are part of their currently active study program.

When a study program of a student is marked as ended in Pyramus, they are automatically removed from active workspaces of that study program only. Earlier implementation would have removed them from every active workspace of their every study program.

Changed workspace permission handling to treat inactive workspace users like they were archived (i.e. permanently deleted). Thus, if a student is removed from a workspace, Muikku will act like they were never even part of it. Naturally they can still sign up for the workspace again, which reactivates the previously inactivated entity.

Renamed most workspace user getters to better indicate that they deal with active workspace users only.

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/WorkspaceBackingBean.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/WorkspaceBackingBean.java
@@ -80,27 +80,25 @@ public class WorkspaceBackingBean {
 
   public void setWorkspaceUrlName(String workspaceUrlName) {
     
-    WorkspaceUserEntity workspaceUserEntity = null;
     WorkspaceEntity workspaceEntity = resolveWorkspaceEntity(workspaceUrlName);
     if (workspaceEntity != null) {
       this.workspaceEntityId = workspaceEntity.getId();
       this.workspaceUrlName = workspaceEntity.getUrlName();
-      if (sessionController.isLoggedIn()) {
-        workspaceUserEntity = workspaceUserEntityController.findWorkspaceUserByWorkspaceEntityAndUserIdentifier(workspaceEntity, sessionController.getLoggedUser());
-      }
     }
-    boolean activeWorkspaceUser = workspaceUserEntity != null && Boolean.TRUE.equals(workspaceUserEntity.getActive());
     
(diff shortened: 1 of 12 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/WorkspaceBackingBean.java
```diff
@@ -79,24 +79,28 @@ public WorkspaceEntity getWorkspaceEntity() {
...
+    
     homeVisible = workspaceToolSettingsController.getToolVisible(workspaceEntity, "home");
     guidesVisible = workspaceToolSettingsController.getToolVisible(workspaceEntity, "guides");
     materialsVisible = workspaceToolSettingsController.getToolVisible(workspaceEntity, "materials");
-    discussionsVisible = sessionController.hasWorkspacePermission(ForumResourcePermissionCollection.FORUM_ACCESSWORKSPACEFORUMS, workspaceEntity) && workspaceToolSettingsController.getToolVisible(workspaceEntity, "discussions");
-    usersVisible = sessionController.hasWorkspacePermission(MuikkuPermissions.MANAGE_WORKSPACE_MEMBERS, workspaceEntity) && workspaceToolSettingsController.getToolVisible(workspaceEntity, "users");
-    journalVisible = sessionController.hasWorkspacePermission(MuikkuPermissions.ACCESS_WORKSPACE_JOURNAL, workspaceEntity) && workspaceToolSettingsController.getToolVisible(workspaceEntity, "journal");
+    discussionsVisible = activeWorkspaceUser && sessionController.hasWorkspacePermission(ForumResourcePermissionCollection.FORUM_ACCESSWORKSPACEFORUMS, workspaceEntity) && workspaceToolSettingsController.getToolVisible(workspaceEntity, "discussions");
```
pkukkon: Doesn't work for administrative users as they fulfill the permission check even when not being workspace users...

PR #3225, opened 2017-06-05
Proper sending and canceling of evaluation requests

### LINKED ISSUE #3224: Unable to cancel evaluation request
If a student has left an evaluation request, pressing the evaluation button claims to cancel that request but instead opens a dialog for leaving another evaluation request.

pkukkon:
- fixes #3224

Method to cancel an evaluation request was there but never called. I added the missing pieces.

While at it, I also removed the "evaluation request canceled" state, because it had no real value and wasn't ever fully implemented either. If an evaluation request is canceled, the resulting workspace state is once again either unassessed or graded.

Furthermore, I also added two new states to indicate a pending evaluation request when the workspace has been graded before. This way, canceling an evaluation request will correctly revert to the previous state of unassessed, passed, or failed.

I also removed the obsolete AssessmentRequestState class, which has been replaced by WorkspaceAssessmentRequestState at some point.

### DIFF
--- a/muikku/src/main/webapp/resources/theme-muikku/sass/_workspaces.scss
+++ b/muikku/src/main/webapp/resources/theme-muikku/sass/_workspaces.scss
@@ -179,7 +179,9 @@ body[class^='workspace'] {
       
     }
     
-    &[data-state='pending'] { /* Evaluation Request Pending */
+    &[data-state='pending'],
+    &[data-state='pending_pass'],
+    &[data-state='pending_fail'] { /* Evaluation Request Pending */
       background:#0099ff; 
       transition: background 0.2s ease-in-out;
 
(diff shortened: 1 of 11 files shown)


muikku/src/main/webapp/resources/theme-muikku/sass/_workspaces.scss
```diff
@@ -179,7 +179,9 @@ body[class^='workspace'] {
       
     }
     
-    &[data-state='pending'] { /* Evaluation Request Pending */
+    &[data-state='pending'],
+    &[data-state='pending_pass'],
+    &[data-state='pending_fail'] { /* Evaluation Request Pending */
```
pkukkon: Backup SCSS; we'll hopefully delete these files altogether at some point.

PR #3365, opened 2017-07-24
Make fonts and paddings smaller and hide some elements

### LINKED ISSUE #2520: Workspace alt(popup) navigation cannot be scrolled
With a very small screen (or very zoomed in) the workspace navigation goes into a small popup menu. However this menu can be so long that it doesn't fit on the screen thus making navigation impossible to perform to the places low in the list.

avukkonen:
Closes #2520 

I did not do the thing asked, because scrolling a popup menu especially in mobile is not the first choice. I removed padding and hid some secondary priority elements out of the way.

### DIFF
--- a/muikku/src/main/webapp/resources/theme-muikku/sass/flex/_workspace_navigation_narrow.scss
+++ b/muikku/src/main/webapp/resources/theme-muikku/sass/flex/_workspace_navigation_narrow.scss
@@ -220,3 +220,228 @@
   }
   
 }
+
+
+
+@media screen and  (max-width: 360px) {
+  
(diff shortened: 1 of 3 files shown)


muikku/src/main/webapp/resources/theme-muikku/sass/flex/_workspace_navigation_narrow.scss
```diff
@@ -219,4 +219,229 @@
...
+            }
+            
+          }
+          
+        }
+        
+        .navmore-evaluation-links {
+          margin-top: 3px;
+          padding: 0;
+          background: #36b497;
```
avukkonen: Not today.

PR #3415, opened 2017-08-17
Bug 3390 frontpage no workspace messages

### LINKED ISSUE #3390: Front page "no workspaces" message issues
Users who are not members of any workspace are shown an instructional message on how to sign up. That message (or rather, the entire Workspaces section) should only be displayed to students, not staff.

Also, the message contains a hyperlink to course picker but it has double quotes and thus leads to a 404 page. Looks like the arg0..* attributes of the localize tag should not use quotes.

jjlankinen:
Closes #3390


muikku/src/main/webapp/index.xhtml
```diff
@@ -212,7 +212,9 @@
 		    <div class="lg-flex-cell12-4 md-flex-cell-8 sm-flex-cell-full lg-flex-cell-first md-flex-order-3 sm-flex-order-3">
 		      <div class="box box-max-width">
 		        <div id="visits"></div> 
-		        <div id="workspaces"></div>
+		        <ui:fragment rendered="#{sessionBackingBean.isStudent}">
```
pkukkon: Changes needed. If staff member has workspaces, the section should render normally. If not, it should not be shown. For students without workspaces, it should render with an instructional message about signing up.

PR #3512, opened 2017-10-06
Ftr 3287 auto create users

### LINKED ISSUE #3287: Auto-create users
#### Details

 - Sync chat users with Pyramus

ilmoeuro:
Resolves #3287


ilmoeuro:
Will not work as-is on production standalone.xml config, need to fix dependency conflict between openfire rest client and wildfly

ilmoeuro:
Forked the rest client library and made it more wildfly-friendly.

PR #3557, opened 2017-11-08
3536 copy workspace image

### LINKED ISSUE #3536: Workspace front page image does not get copied
When making a copy of an workspace, it's front page image does not get copied.

#### OSTicket
http://helpdesk.muikkuverkko.fi/scp/tickets.php?id=2004

#### SH
Nina Kurki

anttiviljakainen:
Closes #3536


muikku-core-plugins/src/main/resources/META-INF/resources/dust/workspacecopywizard/workspace-copy-wizard.dust
```diff
@@ -105,6 +105,21 @@
...
+            <div class="selection-wrapper">
+              <div class="selection-item">
+                <input type="checkbox" id="copy-files" name="copy-files" checked="checked"/>
+              </div>
+              <div class="selection-label">
+                <label for="copy-files" class="selection">{#localize key="plugin.workspacecopywizard.workspaceFiles.label"/}</label>  
+              </div>
+            </div>
+          </div>
+          <div class="flex-cell-8 flex-cell-first">
```
jjlankinen: Should not be flex-cell-first as previous sibling already has flex-cell-first class name indicating it is the first flex item. 

Basically this will do: <div class="flex-cell-8"> ie. just remove flex-cell-first class name.

PR #3573, opened 2017-11-14
Fixed continue studies is off sometimes

### LINKED ISSUE #3501: Frontpage - the "continue studies" navigation is sometimes off.
Student says 

Etusivulla keskellä on kohta "kävit viimeksi työtilassa (esim. ue 1), 
jatka siitä mihin jäit". Se ei toimi. Minulla on ollut siinä sama kohta 
näkyvillä ainakin noin viimeiset 4 vkoa, vaikka olen tehnyt eri kursseja 
lukuisine eri tehtävineen tänä aikana.

http://helpdesk.muikkuverkko.fi/scp/tickets.php?id=1966

onzag:
I found out what it happens, as it was, when you suddenly close the browser, say press the X button because you are done, no other requests are allowed to happen, hence the property setting just doesn't happen as the 'beforeunload' HTTP events doesn't get triggered, even when it could under some circumstances, the browser just wants to close, period.

So with the bugfix the current area will get saved every time the user scrolls to a new section, that way we are avoiding to use it after the beforeunload event, even when I left the event in place; just to keep the old functionality around; nevertheless, it will only get saved as the user scrolls, if the user does not scroll it simply won't happen, so if you navigate to a material screen and you don't scroll it simply won't save. This is the UI behaviour, review code and review the UI behaviour. :)

Fixes #3501

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-materials.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/workspace-materials.js
@@ -1,6 +1,12 @@
 (function() {
   'use strict';
   
+   var timeout = null;
+   //this allows that when one user is scrolling really fast only if a section stays active for more than half
+   //a second it'll be saved, this should be enough so that if the user decides to immediatly close, he/she'll probably
+   //take longer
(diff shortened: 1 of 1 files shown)


onzag:
I think it's fixed now

PR #3745, opened 2018-01-11
Fixes #3427 Workspace diary - saving images in to a diary entry

### LINKED ISSUE #3427: Workspace diary - saving images in to a diary entry
As a student, I would like place an image directly in a diary entry - in addition to the option to bring image from a url. 

http://helpdesk.muikkuverkko.fi/scp/tickets.php?id=1905

RazgrizX:
Fixes #3427 Workspace diary - saving images in to a diary entry
Added plugins to the CKEditor, so it is identical to the one used in Communicator new message.
Warning! : Social navigation is still in use.
Warning:!!! Uses communicatorAttachmentUploadServlet and Communicator Attachment table in database

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/socialnavigation.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/socialnavigation.js
@@ -38,15 +38,22 @@ function openInSN(template, options, formSendFunction, formContentFunction) {
     $(textareas).each(function(index,textarea) {
       $(textarea).val(editableContent);
       if (options && options.draftKey) {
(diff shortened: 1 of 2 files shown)


pkukkon:
Although this implementation would work per se, it was discussed with the team and decided that the image upload functionality in communicator cannot really be reused in journal due to misleading image paths (rest/communicator/etc.) and all images being stored in database as CommunicatorMessageAttachment entities.

There should be a more generic solution to upload and store images, preferably so that any CKEditor field in the application could benefit from it. Once that has been done, the communicator specific image functionality should be refactored to make use of it, too. These have been added as issues #3767 and #3768.

PR #3825, opened 2018-02-15
Null check workspace assessment grade

### LINKED ISSUE #3824: Evaluation REST service fails to handle assessments without grade
NPE due to assuming grading scale and grade always exist

pkukkon:
- fixes #3824


pkukkon:
Check is still insufficient...

PR #3834, opened 2018-02-21
Fixes #3832 xmpp prebind fails

### LINKED ISSUE #3832: Xmpp prebind fails when session closes before converse can reattach to it

saaralepisto:
Fixes #3832 xmpp prebind fails

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatClientHolder.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatClientHolder.java
@@ -1,0 +1,69 @@
+package fi.otavanopisto.muikku.plugins.chat;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.logging.Logger;
+
+import javax.annotation.PostConstruct;
+import javax.ejb.ConcurrencyManagement;
+import javax.ejb.ConcurrencyManagementType;
+import javax.ejb.Schedule;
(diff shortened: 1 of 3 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatClientHolder.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.ejb.ConcurrencyManagement;
+import javax.ejb.ConcurrencyManagementType;
+import javax.ejb.Schedule;
+import javax.ejb.Singleton;
+import javax.inject.Inject;
+
+import rocks.xmpp.core.session.XmppClient;
+
+@Singleton
+@ConcurrencyManagement(ConcurrencyManagementType.BEAN)
```
anttiviljakainen: Is this necessary, iirc the default is write lock anyways so no public method can be called at the same time.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatUserSyncScheduler.java
```diff
@@ -1,112 +1,112 @@
...
-          userEntity = new UserEntity(enabledUser, user.getDisplayName(), "", password);
-          client.createUser(userEntity);
-        }
-      } catch (Exception e) {
-        logger.log(Level.INFO, "Exception when syncing user " + enabledUser, e);
-      }
-    }
-  }
-}
+package fi.otavanopisto.muikku.plugins.chat;
```
anttiviljakainen: Are there some linebreak etc changes that show the whole file as changed?

PR #3839, opened 2018-02-22
Ftr 3772 announcement reading view styling

### LINKED ISSUE #3772: Announcements reading views styling
After templates and functionality is done it's visuals turn.

jjlankinen:
Closes #3772 
This separates announcement view from announcer quite effectively and add new "components". Also fixes the layout to follow the UI sketches more closely.

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/reading-panel.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/reading-panel.tsx
@@ -1,0 +1,112 @@
+import * as React from 'react';
+
+import '~/sass/elements/reading-panel.scss';
+import '~/sass/elements/loaders.scss';
+
+interface ReadingPanelProps {
+  modifier: string,
+  title?: React.ReactElement<any> | string,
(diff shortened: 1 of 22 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/reading-panel.tsx
```diff
@@ -0,0 +1,112 @@
+import * as React from 'react';
```
onzag: If this file was renamed it means that all the other files who used the previous application-panel view are to be broken, they need to be renamed as well.

jjlankinen: Nothing was renamed, reading-panel.tsx was created from application-panel.tsx. Although reading-panel name might need some evaluating as it might be too generic even though that's the idea.

onzag: Ok then no problem I'd guess so, but once I start cleaning up one of them has to go if they represent the same concept, nevertheless, it's good for a pull now.

PR #3841, opened 2018-02-22
Fixes #3840 Select VOPS students based on Pyramus student variable

### LINKED ISSUE #3840: Select VOPS students based on Pyramus student variable

saaralepisto:
Fixes #3840 Select VOPS students based on Pyramus student variable

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/transcriptofrecords/TranscriptOfRecordsController.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/transcriptofrecords/TranscriptOfRecordsController.java
@@ -15,7 +15,6 @@ import javax.inject.Inject;
 import org.apache.commons.lang3.StringUtils;
 
 import fi.otavanopisto.muikku.controller.PluginSettingsController;
-import fi.otavanopisto.muikku.model.users.UserEntity;
 import fi.otavanopisto.muikku.model.workspace.WorkspaceEntity;
 import fi.otavanopisto.muikku.model.workspace.WorkspaceUserEntity;
 import fi.otavanopisto.muikku.schooldata.GradingController;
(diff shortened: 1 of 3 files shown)


muikku-atests/src/test/java/fi/otavanopisto/muikku/TestUtilities.java
```diff
@@ -1,8 +1,11 @@
 package fi.otavanopisto.muikku;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.collections.map.HashedMap;
```
pkukkon: Unused import (Ctrl+Shift+O when editing the source file gets rid of these easily)

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/transcriptofrecords/TranscriptOfRecordsController.java
```diff
@@ -135,19 +135,13 @@ public void saveBoolProperty(User user, String propertyName, boolean value) {
...
-      if (studyViewStudents != null) {
-        String[] ids = studyViewStudents.split(",");
-        for (int i = 0; i < ids.length; i++) {
-          if (StringUtils.equals(ids[i], userEntity.getId().toString())) {
-            return true;
-          }
-        }
-      }
+  public boolean shouldShowStudies(User user) {
+    UserProperty userProperty = userSchoolDataController.getUserProperty(user, "hops.enabled");
```
pkukkon: Apparently this method is used to determine if a view is shown in Muikku, so asking that based on a user variable in Pyramus is slow and rather illogical.

If any user-related information only affects Muikku, that information should be stored in Muikku (e.g. as a UserEntityProperty) as it's faster to retrieve and doesn't clutter Pyramus with data that has no relevance to Pyramus.

anttiviljakainen: Is there a place though in muikku to manage any such properties? I don't see this as major performance problem as (atm) it's used only in ToR and Pyramus has the facilities to manage the property by a user. In the end it's a property that should eventually be removed altogether.

pkukkon: Fiiiiiine... Technically this can probably be left as-is, provided that the variable will eventually become obsolete. At least it follows the same behavior as all other HOPS related variables.

A separate Pyramus issue is probably needed to create the UserVariableKey itself, so that this information can then be managed over there. Irrelevant for this PR, though.

PR #3854, opened 2018-02-28
Fixes #3798 Changing username fails is username is already in use

### LINKED ISSUE #3798: Changing username fails if new username is already in use
Changing credentials has no logic whatsoever to deal with usernames that are already in use (apart that the change itself fails due to a duplicate).

saaralepisto:
Fixes #3798 Changing username fails is username is already in use

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/user/rest/UserPluginRESTService.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/user/rest/UserPluginRESTService.java
@@ -65,10 +65,8 @@ public class UserPluginRESTService extends PluginRESTService {
           userCredentialChange.getUsername(), userCredentialChange.getNewPassword());
     
       return Response.noContent().build();
-    } catch (SchoolDataBridgeUnauthorizedException e) {
-      return Response.status(Status.FORBIDDEN).build();
     } catch (SchoolDataBridgeException e) {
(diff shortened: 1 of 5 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/user/rest/UserPluginRESTService.java
```diff
@@ -66,6 +67,8 @@ public Response updateCredentials(UserCredentials userCredentialChange) {
       return Response.noContent().build();
     } catch (SchoolDataBridgeUnauthorizedException e) {
       return Response.status(Status.FORBIDDEN).build();
+    } catch (SchoolDataBridgeException e) {
+      return Response.status(Status.BAD_REQUEST).build();
```
pkukkon: Since SchoolDataBridgeException contains a (HTTP) statusCode, this catch block and the one above could actually be combined into one:

```
catch (SchoolDataBridgeException e) {
  return Response.status(e.getStatusCode()).build();
}
```

In addition to that, there are two places in `PyramusRestClient` that throw `SchoolDataBridgeInternalException`. If both of them are rewritten as...

`throw new SchoolDataBridgeException(response.getStatus(), String.format("Received http error %d (%s) when requesting %s", response.getStatus(), response.getEntity(), path));`

...then any HTTP error code from any REST call to Pyramus can be passed along in `SchoolDataBridgeException`.

PR #3929, opened 2018-03-21
Fixes #3874 Student can not access announcements

### LINKED ISSUE #3874: Student can not access announcements

RazgrizX:
Added Environmental permission for the user to FIND_ANNOUNCEMENT, since it is checked on the respective REST point and the user need to have access to it in the redesign web site.
Works only after resetting the permissions on https://dev.muikkuverkko.fi:10443/permissions
Note:
Maybe this permission needs to be a workspace permission rather then environmental for the sake of security. This will require changes in authentication on the REST point.


anttiviljakainen:
Cancelling this PR, it gives student permission to find *all* announcements, not just the ones the student is able to see. We'll cover this in another PR.

PR #3931, opened 2018-03-21
Fixes #3877 Announcements for all not showing up

### LINKED ISSUE #3877: Announcer: Announcements for all not showing up.
Steps to reproduce.
1. make announcement
2. edit to have recipient workspace
3. edit not to have recipient workspace

none of the "public" open to all announcements are visible to students anymore.

probably reproducable in other ways too.

RazgrizX:
Fixes #3877 Announcements for all not showing up
Now announcement creator checks the target groups. If there are 0 of them it makes announcement "publiclyVisible"
announcements with at least 1 target group are not "publiclyVisible"


RazgrizX:
If there is no target workspace or group chosen, the this.state.currentTarget.length of the announcer editor is 0 and the announcement should have publiclyVisible parameter set to true. Announcement without target considered public. Previously this parameter been always set to false and it caused all the problems with access. At least, this is how I see this.

PR #3936, opened 2018-03-22
Fixes #3729 preserve chat windows state between page loads

### LINKED ISSUE #3729: Preserve chat windows state between page loads
At the moment chat window minimizes between page loads even if user has specifically opened it for chatting.

saaralepisto:
Fixes #3729 preserve chat windows state between page loads


saaralepisto:
The changes are here:
https://github.com/otavanopisto/converse.js/pull/1

I just had to update the correct paths for flex-main.xhtml

PR #3995, opened 2018-04-18
Fixes and tests to announcements

### LINKED ISSUE #3874: Student can not access announcements

anttiviljakainen:
* Fixed createAnnouncement to actually return the created entity
* Changed permission of finding a single announcement so that students can actually view what they can list

Closes #3874


petja-laitila:
Fixes the problem in redesign.

PR #4066, opened 2018-05-31
FTR #3933 chat opt in and opt out

### LINKED ISSUE #3933: Chat opt-in and opt-out
Users needs to have an option to either opt-in or opt-out of chat feature. This should be placed in profile view.

saaralepisto:
Closes #3933 chat opt in and opt out


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatUserSyncScheduler.java
```diff
@@ -77,22 +84,22 @@ public void updateChatUsers() {
...
-            logger.log(Level.WARNING, "Invalid user identifier " + enabledUser + ", skipping...");
-            continue;
+            logger.log(Level.WARNING, "Invalid user identifier " + listUser.getUserIdentifier() + ", skipping...");
+
           }
           User user = userController.findUserByIdentifier(identifier);
           if (user == null) {
-            logger.log(Level.WARNING, "No user found for identifier " + enabledUser + ", skipping...");
-            continue;
+            logger.log(Level.WARNING, "No user found for identifier " + listUser + ", skipping...");
```
pkukkon: listUser.getUserIdentifier()

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatUserSyncScheduler.java
```diff
@@ -101,12 +108,12 @@ public void updateChatUsers() {
           random.nextBytes(passwordBytes);
           String password = Base64.encodeBase64String(passwordBytes);
 
-          userEntity = new UserEntity(enabledUser, user.getDisplayName(), "", password);
+          userEntity = new UserEntity(listUser.getUserIdentifier(), user.getDisplayName(), "", password);
           client.createUser(userEntity);
         }
       } catch (Exception e) {
-        logger.log(Level.INFO, "Exception when syncing user " + enabledUser, e);
+        logger.log(Level.INFO, "Exception when syncing user " + listUser, e);
```
pkukkon: listUser.getUserIdentifier()

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
```diff
@@ -0,0 +1,68 @@
...
+	
+  private static final long serialVersionUID = -7830619828801454118L;
+
+  public UserChatSettings create(String userIdentifier, UserChatVisibility userChatVisibility) {
+    UserChatSettings settings = new UserChatSettings(userIdentifier, userChatVisibility);
+    getEntityManager().persist(settings);
+    return settings;
+  }
+  
+  public List<UserChatSettings> listAll(String userIdentifier, UserChatVisibility userChatVisibility) {
```
pkukkon: Unused method? All DAO classes have a `listAll` method through class inheritance. This one has `userIdentifier` and `userChatVisibility` as arguments but doesn't use them at all and the code essentially does what plain `listAll()` does, anyway.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
```diff
@@ -0,0 +1,68 @@
...
+	CriteriaQuery<UserChatSettings> criteria = criteriaBuilder.createQuery(UserChatSettings.class);
+	Root<UserChatSettings> root = criteria.from(UserChatSettings.class);
+	criteria.select(root);
+	
+	TypedQuery<UserChatSettings> query = entityManager.createQuery(criteria);
+	
+	return query.getResultList();
+  }
+  
+  public UserChatSettings findByUser(SchoolDataIdentifier userIdentifier) {
```
pkukkon: We try to keep DAO methods as simple and closely resembling the entity class as possible. As such, this method would be better as...

`public UserChatSettings findByUserIdentifier(String userIdentifier)`

...so there's less chance of getting confused whether the method deals with Users, SchoolDataIdentifiers, or a plain String (which it does as the `userIdentifer` in the `UserChatSettings` entity is a string).

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
```diff
@@ -0,0 +1,68 @@
...
+    criteria.select(root);
+    criteria.where(
+        criteriaBuilder.equal(root.get(UserChatSettings_.userIdentifier), userIdentifier.toId())
+    );
+    
+    return getSingleResult(entityManager.createQuery(criteria));
+  }
+  
+  @Override
+  public void delete(UserChatSettings messageSignature) {
```
pkukkon: Unused method. `delete` is another method all DAO classes have through inheritance and even if it has been overridden here, the code simply calls the underlying class so it's not even overriding anything at all.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/profile/ProfileBackingBean.java
```diff
@@ -43,6 +50,12 @@
 
   @Inject
   private LocaleController localeController;
+  
+  @Inject
+  private ChatRESTService chatRESTService;
+  
+  @Inject 
+  PluginSettingsController pluginSettingsController;
```
pkukkon: Missing `private` scope

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/profile/ProfileBackingBean.java
```diff
@@ -99,7 +112,19 @@ public String init() {
     
     SchoolDataIdentifier identifier = new SchoolDataIdentifier(userEntity.getDefaultIdentifier(), userEntity.getDefaultSchoolDataSource().getIdentifier());
     emails = userEmailEntityController.getUserEmailAddresses(identifier);
+   
+    chatEnabled = false;
+    String enabledUsersCsv = pluginSettingsController.getPluginSetting("chat", "enabledUsers");
+    
+    if (enabledUsersCsv == null) {
```
pkukkon: Might as well check with StringUtils.isEmpty(enabledUsersCsv) to deal with both a null or an empty string

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/profile.js
```diff
@@ -1,7 +1,53 @@
...
+            data.visible_to_all_selected = "selected";
+          }
+          if (settings && settings.visibility === "DISABLED"){
+            data.disabled_selected = "selected";
+          }
+          renderDustTemplate('profile/profile-chat-visibility.dust', data, $.proxy(function (text) {
+            this.element.html(text);
+            this.element.find("select").on('change', $.proxy(function(event) {
+              this._setVisibility(event.target.value);
+              setTimeout(function(){ location.reload(); }, 1500);
```
pkukkon: This delayed reload is probably here to show/hide the chat icon? If so, it's better to remove it from here... (see next comment)

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/profile.js
```diff
@@ -1,7 +1,53 @@
...
+            }, this));
+          }, this));
+        }, this));
+      }, this));
+    },
+    
+    _setVisibility: function(visibility) {
+      var userIdentifier = MUIKKU_LOGGED_USER;
+      mApi().chat.settings.update({visibility: visibility, userIdentifier: userIdentifier}).callback($.proxy(function () {
+
```
pkukkon: ...and simply do `location.reload();` in this callback method as it is called as soon as the server-side has stored the new value.

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/profile/ProfileJsPluginMessages_en.properties
```diff
@@ -30,4 +30,8 @@ plugin.profile.changeAddressMunicipality.dialog.municipalityField.label = Munici
 plugin.profile.changeImage.dialog.title = Profile image
 plugin.profile.changeImage.dialog.saveButton.label = Save
 plugin.profile.changeImage.dialog.cancelButton.label = Cancel
-plugin.profile.properties.saved = Information saved
\ No newline at end of file
+plugin.profile.properties.saved = Information saved
+plugin.profile.chat.visibility = Chat visibility
+plugin.profile.chat.visibleToAll = Chat enabled
+plugin.profile.chat.disabled = Chat is disabled
+plugin.profile.chat.visibilityChange = Chat status changet succesfully
```
pkukkon: changed

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/profile/ProfileJsPluginMessages_fi.properties
```diff
@@ -30,4 +30,8 @@ plugin.profile.changeAddressMunicipality.dialog.municipalityField.label = Kotiku
 plugin.profile.changeImage.dialog.title = K�ytt�j�profiilin kuva
 plugin.profile.changeImage.dialog.saveButton.label = Tallenna
 plugin.profile.changeImage.dialog.cancelButton.label = Peruuta
-plugin.profile.properties.saved = Tiedot tallennettu
\ No newline at end of file
+plugin.profile.properties.saved = Tiedot tallennettu
+plugin.profile.chat.visibility = Chat-n�kyvyys
+plugin.profile.chat.visibleToAll = Chat k\u00E4yt\u00F6ss\u00E4
```
pkukkon: Weird encoding for Scandinavian characters. Locale files should be UTF8 so using plain Scandinavian characters should work just fine.

muikku/src/main/webapp/resources/scripts/gui/environment.js
```diff
@@ -1,6 +1,31 @@
 (function() {
   'use strict';
+
+  /* global converse */
   
+  mApi().chat.status.read().callback(function(err, result) {
+    if (result == null || (result && result.visibility == "VISIBLE_TO_ALL")) {
```
pkukkon: Why is `result == null` a passing condition, too? Shouldn't this be `if (result && result.visibility == 'VISIBLE_TO_ALL')` to be sure that chat is only enabled if the status check says so?

muikku-core-plugins-persistence/src/main/java/fi/otavanopisto/muikku/plugins/chat/model/UserChatSettings.java
```diff
@@ -0,0 +1,63 @@
...
+    super();
+    this.userIdentifier = userIdentifier;
+    this.visibility = visibility;
+  }
+
+  public Long getId() {
+    return id;
+  }
+
+  public void setId(Long id) {
```
pkukkon: Hibernate has full control of `id` so never add a setter for it so that no one accidentally uses it.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRESTService.java
```diff
@@ -160,27 +168,44 @@ public Response chatStatus() {
...
+    
+    if (identifier == null) {
+        return Response.status(Status.INTERNAL_SERVER_ERROR).entity("Couldn't find logged user").build();
+      }
+	  
     String enabledUsersCsv = pluginSettingsController.getPluginSetting("chat", "enabledUsers");
+    
     if (enabledUsersCsv == null) {
-      enabledUsersCsv = "";
+      return null;
```
anttiviljakainen: Probably means that the chat is not configured? Likely still better to return something, maybe `Response.status(Status.INTERNAL_SERVER_ERROR).build()` or `Status.NOT_IMPLEMENTED` etc. w/e makes most sense.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRESTService.java
```diff
@@ -160,27 +168,44 @@ public Response chatStatus() {
...
+    if (!enabledUsers.contains(identifier.toId())) {
+      return Response.ok(new StatusRESTModel(false, false, null)).build();
     }
     
+    UserChatSettings userChatSettings = chatController.findUserChatSettings(identifier);
+    
+    if (userChatSettings != null) {
+      UserChatVisibility visibility = userChatSettings.getVisibility();
+    	
+      if (visibility.toString().equals("VISIBLE_TO_ALL")) {
```
anttiviljakainen: `visibility == UserChatVisibility.VISIBLE_TO_ALL` should work and is better to compare two enums.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRoomSyncScheduler.java
```diff
@@ -43,21 +45,25 @@
...
-    List<String> enabledUsers = Arrays.asList(enabledUsersCsv.split(","));
 
     String enabledWorkspacesCsv = pluginSettingsController.getPluginSetting("chat", "enabledWorkspaces");
     if (enabledWorkspacesCsv == null) {
-      return;
+     return;
     }
     List<String> enabledWorkspaces = Arrays.asList(enabledWorkspacesCsv.split(","));
+	  
+	List<UserChatSettings> listUsers = userChatSettingsDao.listAll();
```
anttiviljakainen: `DAO.listAll()` lists all rows from the table. What is the expectation for the row count, does every user have an UserChatSettings, or does just a few users have a row? The synchronization can get too time consuming if there are many rows and in that case the operation needs to be split into chunks. That can be added as another issue though.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
anttiviljakainen: I sort of disagree with the comment because String as parameter is kind of obscure in itself and often having typed parameters helps give correct parameters and enforce type safety via compiler (more of a case when you have f.ex. userEntityId, the DAO method can accept UserEntity as parameter and use UserEntity.getId(), like it's done here).

In this case though the SchoolDataIdentifier is also kind of obscure so I'm fine with w/e.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
anttiviljakainen: IIRC delete is private or protected though so overriding is bumping the visibility. That was the original intent at least so that delete would only be public where designed so.

Variable name could be more suited to the occation though.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/profile/ProfileBackingBean.java
anttiviljakainen: Also to avoid future confusion, move lines 122-126 inside the if i.e.
`if (StringUtils.isNotBlank(enabledUsersCsv) {
  // Lines 122-126
}`
and no `return` clause inside there (breaks the execution of the backing bean init).

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
pkukkon: True dat. Maybe enough to rename method to findByUserIdentifier and leave the argument as-is.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/dao/UserChatSettingsDAO.java
pkukkon: Yeah, sorry, totally forgot delete was protected and has to be made public. In that case, changing messageSignature to userChatSettings is enough.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRESTService.java
```diff
@@ -177,7 +177,7 @@ public Response chatStatus() {
     String enabledUsersCsv = pluginSettingsController.getPluginSetting("chat", "enabledUsers");
     
     if (enabledUsersCsv == null) {
-      return null;
+      Response.status(Status.INTERNAL_SERVER_ERROR).build();
```
anttiviljakainen: Add `return`.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRESTService.java
```diff
@@ -223,5 +248,37 @@ private PrivateKey getPrivateKey() {
...
+	SchoolDataIdentifier userIdentifier = sessionController.getLoggedUser();	  
+	UserChatSettings userChatSettings = chatController.findUserChatSettings(userIdentifier);
+	  
+	return Response.ok(userChatSettings).build();
+  }
+  @PUT
+  @Path("/settings")
+  @RESTPermit(handling = Handling.INLINE)
+  public Response chatSettings(UserChatSettings userChatSettings) {
+	if (!sessionController.isLoggedIn()) {
```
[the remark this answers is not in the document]
anttiviljakainen: Pasi meant the method name, i.e.
`public Response createOrUpdateChatSettings(UserChatSettings userChatSettings) {` :)

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRoomSyncScheduler.java
anttiviljakainen: Was this issue added?

anttiviljakainen:
Check the "show outdated" from the re-review if they don't show automatically.

PR #4096, opened 2018-08-20
Ftr 3968 label dialog styling

### LINKED ISSUE #3968: Label dialog's content's styling is a bit off
When you try to send a new message, modify a label, edit a label, reply to a message... etc... the styles are obviously wrong.

jjlankinen:
Closes #3968


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/link.scss
```diff
@@ -267,6 +267,7 @@
     background-color: $color-main-function-label-link-selected-background;
     
     .text {
```
avukkonen: I know we're getting rid of these .text classes, but - in general - this override I don't like. I'm not sure it's a good practice to override some other components classes. I mean this can be done in a number of ways; Why like this?

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/communicator/CommunicatorJsPluginMessages_en.properties
```diff
@@ -39,6 +39,7 @@ plugin.communicator.label.create.error.alreadyexists = A label with that name al
 plugin.communicator.label.create.error.empty = Name cannot be empty
 plugin.communicator.label.remove.question = Do you want to remove label: "{0}"?
 plugin.communicator.label.edit.caption = Edit label
+plugin.communicator.label.editLabelDialog.name = Name
```
avukkonen: plugin.communicator.label.editLabelDialog.name

Wouldn't this be better like this? :

plugin.communicator.dialog.editLabel.name

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/link.scss
jjlankinen: Only other way to achieve this currently is to add another class name to markup like text--selected-label when either  communicator or guider label link is selected inside label drop down box. That would add in my opinion unnecessary overhead as we need to add corresponding class name to markup and as well add extra css rules to match that class name instead of just override some css rules based on selected state of parent element.

How ever this changes completely when we get rid of text class name.

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/communicator/CommunicatorJsPluginMessages_en.properties
jjlankinen: Can be changed, however there are multiple locale strings currently that are used inside of dialog but are not specified with locale path. We have issue for unifying every locale string we have and I would like to leave this change as it is and take care of it when fixing every locale string.

PR #4177, opened 2018-09-21
Ftr 4122 dialog blocks content

### LINKED ISSUE #4122: Environment dialog blocks content
When user ie. answers to message in communicator the opening environment dialog opens on top of content as planned but currently blocks the content in a way that user can't see ie. the message he/she is answering to.

Legacy version does add necessary padding to the body element and keeps windows scrolling intact to help user to see content underneath the environment dialog.

We need to find some UI/UC solution to this, maybe some sort of hide/show feature that keeps inputted data.

avukkonen:
Closes #4122


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/discussion/dialogs/new-thread.tsx
```diff
@@ -181,7 +181,7 @@ class DicussionNewThread extends SessionStateComponent<DicussionNewThreadProps,
        <div className="env-dialog__row" key="3">     
          <div className="env-dialog__form-element-container">
            <div className="env-dialog__label">{this.props.i18n.text.get('plugin.discussion.createmessage.content')}</div>
-           <CKEditor key="3" width="100%" height="grow" growReference=".env-dialog__body" configuration={ckEditorConfig} extraPlugins={extraPlugins}
+           <CKEditor key="3" width="100%" height="210" growReference=".env-dialog__body" configuration={ckEditorConfig} extraPlugins={extraPlugins}
```
onzag: Modify the ckeditor.tsx file not to use "grow" anymore since after many changes the functionality is broken anyway.

Strip the grow functionality.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/input-contacts-autofill.tsx
```diff
@@ -62,19 +64,37 @@ export default class InputContactsAutofill extends React.Component<InputContacts
...
     this.onDelete = this.onDelete.bind(this);
   }
   componentWillReceiveProps(nextProps: InputContactsAutofillProps){
     if (nextProps.selectedItems !== this.props.selectedItems){
       this.setState({selectedItems: nextProps.selectedItems})
     }
   }
+  
+  
+// Deprecated?
```
onzag: Yes it's not used anywhere, then it should be stripped out of the component.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/input-contacts-autofill.tsx
```diff
@@ -62,19 +64,37 @@ export default class InputContactsAutofill extends React.Component<InputContacts
...
+  setBodyMargin() {
+    let selectedHeight = (this.refs["taginput"] as TagInput).getSelectedHeight();
+    let prevSelectedHeight= this.state.selectedHeight;
+    let currentBodyMargin = parseFloat(document.body.style.marginBottom);       
+    let defaultBodyMargin = currentBodyMargin - prevSelectedHeight;
+    
+    if (selectedHeight !== this.state.selectedHeight){
+      let bodyMargin = defaultBodyMargin + selectedHeight + "px";        
+        document.body.style.marginBottom = bodyMargin;
+        this.setState({selectedHeight: selectedHeight});
```
onzag: Using a setState call causes the component to recheck whether it needs to rerender or recalculate its properties, this can be expensive; you can declare a private number instead the same way there's a NodeJS.Timer private variable in the function, you are not using the selectedHeight to modify the component but rather to modify the body which means the state doesn't really belong to the component; while it works it is cheaper to store this as an attribute of the class rather than an attribute of the state.

So as:

`if (newSelectedHeigth !== this.selectedHeight){`

and

`this.selectedHeight = newSelectedHeight`

If you see the fieldHeight because it was part of the state that's because fieldHeight was used for pixels offset, it affected the component visuals (BTW you need to remove that too since we are dropping setHeight)

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/discussion/dialogs/new-thread.tsx
onzag: Should be stripped as well.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/ckeditor.tsx
```diff
@@ -45,26 +44,7 @@ export default class CKEditor extends React.Component<CKEditorProps, CKEditorSta
...
-        remainingHeight -= parseInt(nComputedStyle.getPropertyValue("height")) +
-          parseInt(nComputedStyle.getPropertyValue("margin-top")) +
-          parseInt(nComputedStyle.getPropertyValue("margin-bottom"));
-      });
-      
-      actualHeight = remainingHeight + ckeditorHeight - 9;
-    } else {
-      actualHeight = height;
-    }
+    actualHeight = height;
```
onzag: actualHeight is not necessary anymore just use height later, if actual height is effectively always equal to height then having two variables with the same value is confusing

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/discussion/dialogs/reply-thread.tsx
```diff
@@ -117,7 +117,7 @@ class ReplyThread extends SessionStateComponent<ReplyThreadProps, ReplyThreadSta
     <div className="env-dialog__row" key="1">     
       <div className="env-dialog__form-element-container">
         <div className="env-dialog__label">{this.props.i18n.text.get('plugin.discussion.createmessage.content')}</div> 
-        <CKEditor autofocus key="1" width="100%" height="grow" growReference=".env-dialog__body" configuration={ckEditorConfig} extraPlugins={extraPlugins}
+        <CKEditor autofocus key="1" width="100%" height="210" growReference=".env-dialog__body" configuration={ckEditorConfig} extraPlugins={extraPlugins}
```
onzag: Let's remove all the growReferences.

PR #4220, opened 2018-10-02
Ftr 3062 redesign of muikku

onzag:
Add the whole of redesign of muikku to the development branch...

This doesn't include the new materials redesign components in the studies view.


muikku/src/main/java/fi/otavanopisto/muikku/session/SessionBackingBean.java
```diff
@@ -84,6 +107,99 @@ else if (user.getNickName() != null) {
...
+    
+    if (sessionController.isLoggedIn()) {
+    	UserEntity userEntity = sessionController.getLoggedUserEntity();
+        User user = userController.findUserByDataSourceAndIdentifier(sessionController.getLoggedUserSchoolDataSource(), sessionController.getLoggedUserIdentifier());
+        List<UserAddress> userAddresses = userController.listUserAddresses(user);
+        List<UserPhoneNumber> userPhoneNumbers = userController.listUserPhoneNumbers(user);
+        
+        displayName = user.getNickName() == null ? user.getDisplayName() : String.format("%s %s (%s)", user.getNickName(), user.getLastName(), user.getStudyProgrammeName());
+        
+        studyStartDate = user.getStudyStartDate();
```
anttiviljakainen: I'm not sure if these fields are used in more generic places than f.ex in guider. If they aren't, they probably should be in some other backing bean (having them in too generic place like this can cause overhead).

muikku-core-plugins/pom.xml
```diff
@@ -328,8 +338,52 @@
 					</execution>
 				</executions>
 			</plugin>
-
-		</plugins>
+			
+			<plugin>
```
anttiviljakainen: Is this plugin needed in daily life? What is it doing?

muikku/src/main/java/fi/otavanopisto/muikku/session/SessionBackingBean.java
anttiviljakainen: Ok. We can look into it later.

muikku-core-plugins/pom.xml
anttiviljakainen: Ok

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/logindetails/LoginDetailsRESTService.java
```diff
@@ -63,5 +64,31 @@ public Response listLogins(@PathParam ("STUDENTIDENTIFIER") String studentId, @Q
     
     return Response.ok(result).build();
   }
-
+  
+  @GET
+  @Path("/students/{STUDENTIDENTIFIER}/loginson")
```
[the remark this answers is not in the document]
anttiviljakainen: It's fine.

muikku-atests/src/test/java/fi/otavanopisto/muikku/ui/AbstractUITest.java
```diff
@@ -737,6 +754,12 @@ public Boolean apply(WebDriver driver) {
     });
   }
   
+  protected void clickOnBullshitElement(String selector) {
```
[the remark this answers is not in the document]
jjlankinen: This will be changed later.

PR #4240, opened 2018-10-19
Scss buidling without ruby

anttiviljakainen:
Scss files are compiled under target/scss-compiled and from there copied to war with the war plugin in pom.xml.

target/scss-compiled can be used in eclipse by adding it to deployment assembly deploying to 'resources' path


pkukkon:
Outdated or dev environment dependent?

PR #4265, opened 2018-11-07
Ftr 4260 progressbar for studies

### LINKED ISSUE #4260: Create progressbar for studies view
Create a progressbar for studies as designed and seen in invisionapp

avukkonen:
closes #4260


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/records/body/application/records.tsx
```diff
@@ -111,47 +107,15 @@ function getActivity(props: RecordsProps, workspace: WorkspaceType){
...
-            left: workspace.studentActivity.exercisesDonePercent === 0 ? "0%" : null,
-            right: workspace.studentActivity.exercisesDonePercent === 0 ? null : 100 - workspace.studentActivity.exercisesDonePercent + "%"
-          }
-        }
-      }}
-      strokeWidth={1} easing="easeInOut" duration={1000} color="#ff9900" trailColor="#f5f5f5"
-      trailWidth={1} svgStyle={{width: "100%", height: "4px"}}
-      text={workspace.studentActivity.exercisesAnswered + "/" + workspace.studentActivity.exercisesTotal}
-      progress={workspace.studentActivity.exercisesDonePercent/100}/> : null}
+    return <div className="application-list__header-secondary">
```
jjlankinen: Can this be header-secondary as it will be placed after header-primary with other "secondary" elements like evaluation request and assessment elements?

In the end we end up in the following markup with multiple secondaries:

`<div class="application-list__header">`
`<div class="application-list__header-primary">...</div>`
`<div class="application-list__header-secondary">...</div>`
`<div class="application-list__header-secondary">...</div>`
`<div class="application-list__header-secondary">...</div>`
`</div>`

Is this desirable to have it this way? Maybe we should place all those elements inside of header-secondary elements and name them individually?

This has been the case earlier before this activity badge as well so not really needed to be changed here now.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/activity-badge.scss
```diff
@@ -0,0 +1,50 @@
...
+  }
+}
+
+.activity-badge__unit-bar {
+  background-color: $color-default-contrast;
+  width:100%;
+}
+
+@for $i from 1 through $progress-units {
+    $height: $i + %;
```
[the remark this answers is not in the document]
avukkonen: It is the only way it will work, dear karvatura

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/transcriptofrecords/TranscriptofRecordsJsPlugin_fi.properties
```diff
@@ -1,72 +1,67 @@
...
-plugin.records.summary.card.activity.title = Aktiivisuus
-plugin.records.summary.card.activity.stat.pre = Olet kirjautunut yhteens�
-plugin.records.summary.card.activity.stat.post = kertaa viimeisen kuukauden aikana
-plugin.records.summary.card.tasks.title = Palautetut teht�v�t
-plugin.records.summary.card.tasks.stat.pre = Olet palauttanut yhteens�
-plugin.records.summary.card.tasks.stat.post = arvioitavaa teht�v�� opiskelujesi aikana
-plugin.records.summary.notifications.title = Ilmoitukset
-
-
+plugin.records.records.title = Records
```
jjlankinen: EN locale strings in FI properties file? 

Seems like there are mixed locale strings in FI and EN locale files in general regarding records view.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/activity-badge.scss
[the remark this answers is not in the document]
jjlankinen: We should use percentage function provided by sass instead of defining number as a string when interpolating as we can't add the unit directly after $i variable. Hound is complaining about failed calculation basically. If you add **1px** then its a calculation and if you add **bold** its transformed as a string like mentioned in the e.g part. 

If you do it like this `$height: $i + %` the $height will be a string that looks like a number instead of a number.

This should be the correct way according the interwebs:
```
$progress-units: 100;

@for $i from 1 through $progress-units {
  
  .activity-badge__unit-bar--#{$i} {
    height: percentage($i/100);
  }
}
```

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/records/body/application/records.tsx
avukkonen: Yeah, I noticed it too. The secondary should ideally contain all of the badges.

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/transcriptofrecords/TranscriptofRecordsJsPlugin_fi.properties
avukkonen: Fucking 3 finger death paste

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/records/body/application/records.tsx
avukkonen: Well, it's fixed here and now

PR #4313, opened 2018-11-30
4304 recipients

### LINKED ISSUE #4304: Clarify recipient search

anttiviljakainen:
Closes #4304

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/rest/CommunicatorRecipientsRESTService.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/rest/CommunicatorRecipientsRESTService.java
@@ -44,7 +44,6 @@ import fi.otavanopisto.muikku.search.SearchProvider;
 import fi.otavanopisto.muikku.search.SearchProvider.Sort;
 import fi.otavanopisto.muikku.search.SearchResult;
 import fi.otavanopisto.muikku.security.RoleFeatures;
-import fi.otavanopisto.muikku.servlet.BaseUrl;
 import fi.otavanopisto.muikku.session.SessionController;
 import fi.otavanopisto.muikku.users.EnvironmentUserController;
 import fi.otavanopisto.muikku.users.UserEmailEntityController;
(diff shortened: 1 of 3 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/rest/CommunicatorRecipientsRESTService.java
```diff
@@ -0,0 +1,308 @@
...
+  @Inject
+  private WorkspaceUserEntityController workspaceUserEntityController;
+  
+  @Inject
+  @Any
+  private Instance<SearchProvider> searchProviders;
+
+  @GET
+  @Path("/recipientsUsersSearch") // TODO mApi requires ids between all resources - this should be /recipients/users/search
+  @RESTPermitUnimplemented
```
pkukkon: Might be more accurate as...
`@RESTPermit(handling = Handling.INLINE, requireLoggedIn = true)`
...as it also does away with the logged in check at the beginning of the method

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/rest/CommunicatorRecipientsRESTService.java
```diff
@@ -0,0 +1,308 @@
...
+          firstResult, 
+          maxResults);
+      
+      List<Map<String, Object>> results = result.getResults();
+
+      List<fi.otavanopisto.muikku.rest.model.User> ret = new ArrayList<fi.otavanopisto.muikku.rest.model.User>();
+
+      for (Map<String, Object> o : results) {
+        String[] id = ((String) o.get("id")).split("/", 2);
+        UserEntity userEntity = userEntityController
```
pkukkon: Users in the search index also have userEntityId so could be slightly simpler to find with that?

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/rest/CommunicatorRecipientsRESTService.java
```diff
@@ -0,0 +1,308 @@
...
+            (String) o.get("nickName"), 
+            hasImage,
+            (String) o.get("nationality"),
+            (String) o.get("language"), 
+            (String) o.get("municipality"), 
+            (String) o.get("school"), 
+            emailAddress,
+            studyStartDate,
+            studyTimeEnd));
+        }
```
pkukkon: Worth logging a warning in else block? Users found in search index but not in database is something that shouldn't really happen (but sometimes does, especially when fooling around with dev environment).

PR #4372, opened 2019-01-03
Remove base64 encoded audio clips

### LINKED ISSUE #4350: Remove embedded audio clips

pkukkon:
- removes audio clips embedded into html in base64 encoded flac and wav flavors that cause massive overhead
- a shoddy rewrite that only supports flac without wav fallback
- complete redesign required
- resolves #4350

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/audiorecord.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/audiorecord.js
@@ -9,7 +9,7 @@
       leftChannel: false,
       disableLogs: false,
       maxClipLength: 60000 * 5,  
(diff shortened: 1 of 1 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/gui/audiorecord.js
```diff
@@ -624,29 +674,26 @@
...
-            type: type,
-            blob: file
-          }
+          name: name,
+          type: type,
+          blob: file
         });
-      } else if (type == 'audio/flac') {
+      }
+      else if (type == 'audio/flac') {
```
anttiviljakainen: Same content under if as above

PR #4423, opened 2019-01-31
FTR-4422 Download audio icon

### LINKED ISSUE #4422: Download icon for along side audio recordings
We should have universal download link added next to the audio field (this is created after audio recording is successful) so every browser can download the recording as browsers native audio player might not offer this feature natively.

jjlankinen:
Fixes #4422


jjlankinen:
Hound errors has been resolved incorrectly but there is no point to fix them  ow as we are currently rewriting everything relating to these changes anyways.

PR #4424, opened 2019-01-31
Added message edit state button

### LINKED ISSUE #4394: Unread button missing in communicator message reading view
The button to set message to unread state is missing in message reading view.

onzag:
Fixes #4394


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/communicator/body/application/toolbar.tsx
```diff
@@ -144,6 +165,8 @@ class CommunicatorToolbar extends React.Component<CommunicatorToolbarProps, Comm
             }>
               <ButtonPill buttonModifiers="label" icon="tag"/>
             </Dropdown>
+            {isUnreadOrInboxOrLabel ? <ButtonPill buttonModifiers="toggle-read" icon={`message-${this.state.isCurrentRead ? "" : "un"}read`}
```
jjlankinen: This is ok for now but consider if icons name changes from icon-read/icon-unread to something completely different. Then we can't just prefix the icon suffix that just prefixes read to unread. Might be better to actually handle both cases separately so unread and read icon names are not related to each other.

PR #4434, opened 2019-02-04
Ftr 4404 matriculation subject eligibility redux

### LINKED ISSUE #4404: Matriculation subject eligiblility reduxification
Theres' still the box that Metatavu made that's a bigger component and needs to be reduxified.

avukkonen:
Closes #4404


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/main-function/records/subject_eligibility.ts
```diff
@@ -0,0 +1,39 @@
...
+          acceptedCount: subjectEligibility.acceptedCourseCount + subjectEligibility.acceptedTransferCreditCount,
+          loading: false
+      }
+      dispatch({
+        type: 'UPDATE_STUDIES_SUBJECT_ELIGIBILITY',
+        payload: subjectEligibilityData
+      });
+    }
+    catch(err) {
+      //TODO: ERR
```
onzag: TODO? does this go to production?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/records/body/application/matriculation-eligibility-row/matriculation-eligibility-row.tsx
```diff
@@ -114,20 +66,21 @@ class MatriculationEligibilityRow extends React.Component<MatriculationEligibili
...
   /**
    * Returns list text for student matriculation eligibility
    * 
    * @returns list text for student matriculation eligibility
    */
   getEligibleText(): string {
-    switch (this.state.eligible) { 
-      case EligbleEnum.TRUE:
+    switch (this.props.subjectEligibility.egilibility) { 
+      case "TRUE":
```
onzag: Not necessary but we could use something else than TRUE or FALSE, maybe, ELEGIBLE and NOT_ELEGIBLE, because these are strings, I know that we have a UNKNOWN case, but it makes it that it's not very clear why we didn't use the native boolean value, and makes it hard to read the code later when we didn't know what this all was about.

PR #4572, opened 2019-04-09
Changed how the mandatory courses are calculated

### LINKED ISSUE #4571: Matriculation calculation issues

anttiviljakainen:
* Changed highlighting from whole row to just cells in enrolled list
* Added highlighting of conflicting mandatory cells (when exam is in finished list and is being repeat)
* Added highlighting of repeat cell when the same subject is in finished list (ie implied repeat)
* Added warning messages of the above two
* Changed how most of the counts are taken - previously Improbatur grades were the only ones excluded, but now duplicates of a subject are removed if they come from the finished list, allowing repeat mechanism to work despite the given grade (needs to work this way)

Closes #4571

### DIFF
--- a/matriculation-plugin/src/main/resources/META-INF/resources/scripts/gui/matriculation/matriculation.jsx
+++ b/matriculation-plugin/src/main/resources/META-INF/resources/scripts/gui/matriculation/matriculation.jsx
@@ -550,6 +550,9 @@ class App extends React.Component {
             + date.getFullYear(),
       lastSave: date.getTime()
     };
+    
+    this.isConflictingMandatory = this.isConflictingMandatory.bind(this);
+    this.isConflictingRepeat = this.isConflictingRepeat.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


matriculation-plugin/src/main/resources/META-INF/resources/scripts/gui/matriculation/matriculation.jsx
```diff
@@ -976,8 +1040,12 @@ class App extends React.Component {
                   amountOfMandatoryAttendances={ this.getAmountOfMandatoryAttendances() }
                   amountOfAcademicSubjectAttendances={ this.getAmountOfAcademicSubjectAttendances() }
                   amountOfMandatoryAdvancedSubjectAttendances = { this.getAmountOfMandatoryAdvancedSubjectAttendances() }
+                  isConflictingMandatory={(attendance) => { return this.isConflictingMandatory(attendance);}}
                   conflictingAttendances={this.isConflictingAttendances()}
+                  isConflictingRepeat={(attendance) => { return this.isConflictingRepeat(attendance);}}
```
onzag: you should pass `this.isConflictingRepeat` alone as it is basically equivalent

If you are having trouble with context being undefined you need to do

`this.isConflictingRepeat = this.isConflictingRepeat.bind(this)` at constructor time, this is an issue from react that comes from javascript changing contexts.

onzag: The same applies to the other attributes written in a similar fashion.

PR #4865, opened 2020-01-17
Sync transcript of records with redesign

### LINKED ISSUE #4861: Sync transcript of records and remove deprecated files with redesign

onzag:
Fixes #4861 

Server and client side file synced, due to the sheer amount of conflicting data the strategy used was not to get any errors and get as close as possible, some issues that might have arised:

1. Class names reverted. (Check design)
2. Sections of code stripped during the merge (Check functionality mainly in Cousepicker and Studies View)
3. Type conflict between server and client side (eg. client expects some data and server provides another data, this would be visible as empty fields where there should have been a value).

Please check each and if you find anything, if you are able to fix them yourself, please do so against this PR, otherwise, refer to me.

I'd assume @jjlankinen would be able to spot design issues, @pkukkon server issues, and @petja-laitila would be able to test things for missing data or sections of code.

If a section of code is missing, we should be able to retrieve back easily by checking with dev.

This branch will ensure things are sync between redesign-dev and any other branch to keep a consistent base. Note that there is quite a bit of dead code in the reducers (that is future code that will come in the redesign and it was the simplest solution to apply even if it doesn't do anything currently).


muikku-core-plugins/src/main/resources/META-INF/resources/jsf/records/index.xhtml
```diff
@@ -5,35 +5,26 @@
...
-  	  FILES = #{transcriptofRecordsBackingBean.files};
-  	  STUDY_START_DATE = "#{transcriptofRecordsBackingBean.studyStartDate}";
-  	  STUDY_TIME_END = "#{transcriptofRecordsBackingBean.studyTimeEnd}";
-  	  STUDY_END_DATE = "#{transcriptofRecordsBackingBean.studyEndDate}";
-    </script>
     <script defer="defer" jsf:name="scripts/dist/vendor.js" />
-    <script defer="defer" jsf:name="scripts/dist/records.js" />
+    <script defer="defer" jsf:name="scripts/dist/main-function.js" />
+    
+<!--     /* MUIKKU_PERMISSIONS.SHOW_STUDIES = #{transcriptofRecordsBackingBean.showStudies}; */ -->
```
anttiviljakainen: Remove the comments from this file?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/guider/body/application/current-student/workspaces/workspace.tsx
```diff
@@ -120,23 +118,23 @@ class StudentWorkspace extends React.Component<StudentWorkspaceProps, StudentWor
...
+
     let stateText;
     let extraClasses = "";
     switch (workspace.studentActivity.assessmentState.state){
       case "pending":
       case "pending_pass":
       case "pending_fail":
         stateText = "plugin.guider.assessmentState.PENDING";
-        extraClasses = "state-PENDING";
+        extraClasses = "state-pending";
```
jjlankinen: Shouldn't we use caps here as in other cases too?

In current-record.tsx file we use state-PASSED etc. but here we use lower case.

PR #4935, opened 2020-04-27
Communicator trash folder rewrite

### LINKED ISSUE #4929: Communicator trash folder performance
With a high enough number of messages (in general, not user's own), it seems communicator trash folder takes quite some time to load.

pkukkon:
- resolves #4929


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/communicator/dao/CommunicatorMessageDAO.java
```diff
@@ -213,75 +213,68 @@ public CommunicatorMessageId findNewerThreadId(UserEntity userEntity, Communicat
...
+        return result == 0 ? o2.getCreated().compareTo(o1.getCreated()) : result;
+      }
+    };
+    trashedMessages.sort(comparator);
+    
+    // Filter to only the latest message in each thread
+    
+    if (!trashedMessages.isEmpty()) {
+      int i = 0;
+      while (true) {
```
anttiviljakainen, line 260: maybe `while (i < trashedMessages.size())` and remove the if with break inside the loop, at least looks safer than the scary while(true) :)

PR #5178, opened 2020-11-05
Chat functionality

### LINKED ISSUE #4564: Chat redesign
Goal is to redesign chat functionality using headless ConverseJS and build UI and front end using React. Chat server remains the same.

Parent of #4590 #4716 #4719 #4723 #4741 #4754 #4757 #4767

pkukkon:
Closes #4564


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/webpack.config.js
```diff
@@ -81,7 +81,8 @@ module.exports = {
 	entry: entries,
 	output: {
 		filename: "[name].js",
-		path: __dirname + "/../dist"
+		path: __dirname + "/../dist",
+		jsonpFunction: "webpackjsonpFunction3"
```
onzag, line 85: Where did this come from?....
Does not seem to affect anything nevertheless since there's no jsonp usage.
Maybe we should consider revert?...

jjlankinen: Seems to be coming from this commit: https://github.com/otavanopisto/muikku/commit/bff6985452453019fa6a36a4e070cf7234eaf4b7

It was needed when we compiled our front end with 3rd party library (we needed to use webpack3 then) and it's not relevant anymore as we have gotten rid of converseJs.

Should be reverted in my opinion.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
```diff
@@ -671,7 +676,17 @@ let updateWorkspace:UpdateWorkspaceTriggerType = function updateWorkspace(data){
         let additionalInfo  = <WorkspaceAdditionalInfoType>(await promisify(mApi().workspace.workspaces.additionalInfo.cacheClear().read(currentWorkspace.id), 'callback')());
 
         data.update.additionalInfo = additionalInfo;
+      }
 
+      // Update workspace chat status (enabled/disabled)
+      if (newChatStatus) {
```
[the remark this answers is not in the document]
jjlankinen: Above comment seems to be out of place and made to the wrong place. Line 682 clearly takes care of updating workspace chat status.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
```diff
@@ -267,47 +267,51 @@ let setCurrentWorkspace:SetCurrentWorkspaceTriggerType = function setCurrentWork
...
 
-                                                 (data.loadDetails || workspace && workspace.details) ? reuseExistantValue(true, workspace && workspace.details,
-                                                   ()=>promisify(mApi().workspace.workspaces
-                                                       .details.read(data.workspaceId), 'callback')()) : null,
+                                                (data.loadDetails || workspace && workspace.details) ? reuseExistantValue(true, workspace && workspace.details,
+                                                  ()=>promisify(mApi().workspace.workspaces
+                                                    .details.read(data.workspaceId), 'callback')()) : null,
 
-                                                  ]) as any
+                                                reuseExistantValue(true, workspace && workspace.chatStatus,
```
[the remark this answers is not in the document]
pkukkon: Both GET (mApi().chat.workspaceChatSettings.read) and PUT (mApi().chat.workspaceChatSettings.update) endpoints exist in ChatRESTService and looks like this file calls them correctly.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/webpack.config.js
pkukkon: Glad someone knew what it was :D All for revert as well

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
```diff
@@ -267,47 +267,51 @@ let setCurrentWorkspace:SetCurrentWorkspaceTriggerType = function setCurrentWork
       let isCourseMember:boolean;
       let journals:WorkspaceJournalsType;
       let details:WorkspaceDetailsType;
+      let chatStatus: WorkspaceChatStatusType;
       let status = getState().status;
-      [workspace, assesments, feeInfo, assessmentRequests, activity, additionalInfo, contentDescription, producers, isCourseMember, journals, details] = await Promise.all([
-                                                 reuseExistantValue(true, workspace, ()=>promisify(mApi().workspace.workspaces.cacheClear().read(data.workspaceId), 'callback')()),
+      [workspace, assesments, feeInfo, assessmentRequests, activity, additionalInfo, contentDescription, producers, isCourseMember, journals, details, chatStatus] = await Promise.all([
+                                                reuseExistantValue(true, workspace, ()=>promisify(mApi().workspace.workspaces.cacheClear().read(data.workspaceId), 'callback')()),
```
pkukkon, line 273: Indent of 48 is a bit excessive, eh?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/workspaces/index.ts
```diff
@@ -357,7 +359,7 @@ export interface OrganizationWorkspacesType {
   activeFilters: OrganizationWorkspacesActiveFiltersType,
   hasMore: boolean,
   toolbarLock: boolean,
-  types?: Array<WorkspaceTypeType>
+  types?: Array<WorkspaceTypeType>,
```
pkukkon, line 362: Why comma after last element?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/workspaces/index.ts
```diff
@@ -373,13 +375,13 @@ export interface MaterialAssignmentType {
   assignmentType: "EXERCISE" | "EVALUATED",
   correctAnswers: string,
   path: string,
-  title: string
+  title: string,
```
pkukkon, line 378: Why comma after last element?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/workspaces/index.ts
```diff
@@ -373,13 +375,13 @@ export interface MaterialAssignmentType {
...
 }
 
 export interface MaterialContentNodeProducerType {
-  id: number;
-  name: string;
-  materialId: number;
+  id: number,
+  name: string,
+  materialId: number,
```
pkukkon, line 384: Why comma after last element?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/workspaces/index.ts
jjlankinen: Trailing commas in type/array/object literals are more of a styling preference. We might as well decide in general whether we use them or not. I took habit of adding them which is or is not the preferred way.

We could even add rule about trialing commas to the hound's typescript linter and fix them one by one eventually.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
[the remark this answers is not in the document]
onzag: Right it seems I misinterpreted the diff.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
onzag: Self formatting... I think we shouldn't worry about it now, it just happened when using eclipse.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/workspaces/index.ts
onzag: Normally it's preferred to be a ";" but when we developed muikku using eclipse we used commas because there was no style preference then.

The purpose of adding a , at the end is that when diffing later if you add one last element, the line where a comma exists will not be affected

{
  propertyA: string,
  propertyB: string,
}

Then we need to introduce propertyC

{
  propertyA: string,
  propertyB: string,
  propertyC: string,
}

Now when passing through diffing tool only propertyC line will be highlighted as the only difference.

If there wasn't a comma this wouldn't happen; that's the reason its' preferred; now it should usually be ";" but we used some old style when making muikku when we used eclipse then.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/workspaces/index.ts
onzag: Same reason, for diffing purposes and a stylistic preference.

Now ";" is preferred but we used comma when making muikku because that was what we had then, and it's kept that way.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatSyncController.java
```diff
@@ -58,279 +78,377 @@
...
+  private void addAdminsAsRoomOwners(RestApiClient client, MUCRoomEntity mucRoomEntity) {
+    SearchProvider searchProvider = getSearchProvider();
+    if (searchProvider == null) {
+      logger.severe("ElasticSearch missing");
       return;
     }
-    if (!StringUtils.isNumeric(openfirePort)) {
-      logger.log(Level.WARNING, "Invalid openfire port, skipping room sync");
-      return;
+    List<OrganizationEntity> organizations = organizationEntityController.listLoggedUserOrganizations();
```
anttiviljakainen, line 413: Is using logged user here a correct call?

pkukkon: We have to limit the search to return admins and study programmer leaders of only "our own" organization, so the code assumes the organization is that of the logged-in user doing stuff that causes this method to be called (basically whenever a new chat room is created).

anttiviljakainen (approved):

PR #5206, opened 2020-11-26
Minor chat bug fixes

pkukkon:
- permission check fix when updating workspace chat settings
- 403 fixes when asking chat status when not logged in


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/chat/ChatRESTService.java
```diff
@@ -503,6 +508,10 @@ public Response createOrUpdateWorkspaceChatSettings(@PathParam("WorkspaceEntityI
       return Response.status(Status.BAD_REQUEST).build();
     }
 
+    if (!sessionController.hasWorkspacePermission(MuikkuPermissions.MANAGE_WORKSPACE_DETAILS, workspaceEntity)) {
```
petja-laitila: This could be the first thing to do. To avoid unnecessary calls.

anttiviljakainen (approved):

PR #5217, opened 2020-12-09
Groupchat moderation

### LINKED ISSUE #5194: Chat message moderation (edit/delete) in MUCs
User should be able to edit and delete own messages. 
Staff member should be able to edit/delete any message.

This feature set is part of MUC's only, does not apply to private messages. 

If message is deleted it will be replaced with another message that says "Message deleted". If message is edited it will be replaced with the another message.

We will not follow specs of XMPP fully as message moderation if not even an draft yet. Instead we create our own client support for moderation as we have full control of users clients.

pkukkon:
closes #5194


muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/chat/ChatJsPluginMessages_en.properties
```diff
@@ -19,12 +19,27 @@ plugin.chat.occupants.students = Student
...
-plugins.chat.notification.createFail = Creating chat room failed, please try agian.
\ No newline at end of file
+plugin.chat.messages.messageIsDeleted = Message deleted
+plugin.chat.messages.deleteMessage = Delete
+plugin.chat.messages.editMessage = Edit
+plugin.chat.messages.editMessage.cancelLink = Cancel
+plugin.chat.messages.editMessage.orText = or
+plugin.chat.messages.editMessage.saveLink = Save
+plugin.chat.messages.deleteMessageTitle = Delete message
+plugin.chat.messages.deleteMessageDesc = Are you absolutely sure you want to delete message?
```
avukkonen, line 36: "delete the message"

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/chat/ChatJsPluginMessages_en.properties
```diff
@@ -19,12 +19,27 @@ plugin.chat.occupants.students = Student
...
+
+plugin.chat.messages.edited = Edited
+
+plugins.chat.notification.roomDeleteSuccess = Chat room <b>{0}</b> was succesfully deleted.
```
avukkonen, line 40: These could all use "the".

avukkonen (changes requested):

avukkonen (approved):

PR #5243, opened 2021-01-13
Upgrade to webpack 5

onzag:
This PR basically upgrades webpack to webpack 5 for a small optimization boost.

There's not much to be added to it to add speed, as parallel webpack proved not to help and it webpack uses only 1 out of the 8 cores, worse in single thread mode.

Some changes to the JSF were required.

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/webpack.config.js
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/webpack.config.js
@@ -33,7 +33,11 @@ rules.push({
 		},
 		{
 			loader: "css-loader",
-			options: { sourceMap: true, importLoaders: 1 },
+			options: {
+				sourceMap: true,
+				importLoaders: 1,
+				url: false
+			},
 		},
 		{
 			loader: "sass-loader",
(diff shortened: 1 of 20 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/webpack.config.js
```diff
@@ -2,83 +2,76 @@ let fs = require('fs');
...
 
-let ExtractTextPlugin = require('extract-text-webpack-plugin');
+const isDevelopment = process.env.NODE_ENV !== "production";
+const mode = isDevelopment ? "development" : "production";
 
-let plugins = [
-	new webpack.optimize.CommonsChunkPlugin({
-		name: 'vendor',
-		minChunks: ({ context }) => context && context.includes("node_modules"),
+let MiniCSSExtractPlguin = require('mini-css-extract-plugin');
```
pkukkon, line 8: Typo? Should probably be MiniCSSExtractPlugin

pkukkon (changes requested):

pkukkon (dismissed):

pkukkon (approved):

PR #5247, opened 2021-01-15
Fix dropdown component not to overflow

### LINKED ISSUE #4956: Dropdown component's tooltip overflows mobile devices screen
If select task field has example text added (which is shown in exercises after user checks exercise) the tooltip often renders outside of mobile devices viewport.

onzag:
Fixes #4956 

Based it on the webpack 5 branch which I mistakenly assumed was already merged.

I could cherrypick the changes specific to the dropdown or we could just wait until that one is merged.


pkukkon (approved):

jjlankinen (approved):
Does what it's suppose to do, locally tested also.

PR #5263, opened 2021-01-25
scrollbar hyperlinks

### LINKED ISSUE #4958: Hyperlinks to material pages don't work reliably
Browsers seem to prevent/adjust users scroll position quite aggressively and our implementation currently don't work reliably when creating hyperlinks to single pages.

Desired behavior would be that when user click the hyperlink pointing to a single page in the middle of workspace materials the browser would scroll precisely to that position. Somehow basic anchors does work but at least Chrome has been reported to override those as well in the past.

Nevertheless we need to find some sort of solution to this as in production using the old tech this works 99% of times but in Redesign it seem to never work reliably.

onzag:
Closes #4958 #5037

I believe so it really needs testing, after trying many solutions from custom scrollbars to other methods, this extremely aggressive scrolling mechanism is what I set to.

We need to ensure it doesn't become an infinite loop because that is what it is.


jjlankinen (approved):
Like discussed before this is brute force scrolling but the only way basically to force scroll position accurately because of browsers intelligent (read: disastrously stupid) scrolling. 

Seal of approval!

jjlankinen (dismissed):

jjlankinen (approved):

PR #5290, opened 2021-02-03
search field clearing functionality

### LINKED ISSUE #5260: Search fields needs functionality to clear search criteria
Search fields used in guider, communicator, course picker, workspace management etc. really could use some sort of search criteria clearing functionality. 

Currently user has to manually delete the search string which is a bit "old" way. We should provide icon that will act as a trigger to remove the search criteria.

avukkonen:
Closes #5260


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/form-element.tsx
```diff
@@ -160,23 +164,28 @@ export class SearchFormElement extends React.Component<SearchFormElementProps, S
...
-        <div className="form-element__input-decoration--main-function-search icon-search"></div>
+      <div className={`form-element form-element--search ${this.props.modifiers ? modifiers.map(m => `form-element--${m}`).join(" ") : ""} ${this.props.className ? this.props.className : ""}`}>
+        <label htmlFor={this.props.id} className="visually-hidden">{this.props.placeholder}</label>
+        <input id={this.props.id} onFocus={this.props.onFocus} onBlur={this.props.onBlur} name={this.props.name} value={this.props.value} className={`form-element__input form-element__input--search ${this.props.modifiers ? modifiers.map(m => `form-element__input--${m}`).join(" ") : ""}`} placeholder={this.props.placeholder} onChange={this.updateSearchField} />
+        <div className={`form-element__input-decoration--clear-search icon-cross ${this.props.value.length > 0 ? "active" : ""}`} onClick={this.clearSearchField}></div>
+        <div className="form-element__input-decoration--search icon-search"></div>
```
jjlankinen, line 188: We should transform **search** and **clear search** elements to buttons  in the future and add visuaally-hidden labels to them so we can then provide support for keyboard users as well within search fields. If we use buttons instead of div elements then keybords tabindex will work automatically, hidden label with proper `id` and `htmlFor` attributes then provides information to the screen readers as well.

Also `icon-search` element could have some sort of event tied to it so it can actually work as an search button as well even if we don't require the button press to activate search currently. We need to investigate what accessibility directive specifies in those cases where search is not trigger via button press but instead is automatic based on users typing intervals.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/base/mixins.scss
```diff
@@ -102,6 +102,22 @@
...
+  align-items: center;
+  color: $color-search-clear-icon;
+  display: flex;
+  font-size: 0.6rem;
+  height: 1.8rem;
+  justify-content: center;
+  width: 1.8rem;
+
+  @include breakpoint($breakpoint-pad) {
+    font-size: 0.6rem;
```
jjlankinen, line 115: `font-size`, `height` and `width` declarations does not seem necessary here as they don't change at all from the "root" values.

jjlankinen (changes requested):

jjlankinen (approved):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/form-element.tsx
```diff
@@ -257,7 +277,7 @@ interface SelectFormElementProps {
   modifiers?: string | Array<string>,
   updateField: (value: string, valid: boolean, name: string) => any;
 }
-
+1
```
onzag, line 280: I think this doesn't belong here.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/form-element.tsx
```diff
@@ -147,36 +146,57 @@ export class FormWizardActions extends React.Component<FormWizardActionsProps, F
...
     let value = e.target.value;
+    if (e.target.value.length > 0) {
+      this.setState({ clearActive: true });
+    } else {
+      this.setState({ clearActive: false });
+    }
     this.props.updateField(value);
   }
 
+  clearSearchField() {
```
onzag, line 184: I checked the codebase and it seems that the property value is never used and undefined is passed every time which means that the field is always uncontrolled, despite making it seem to be controlled.

Also the clearActive state serves no purpose, what I think should be done instead could be for example.

1. Remove clear active, and calculate whether the button should be active whether we have a value or not depending to what is provided.
2. Remove value from the props, and add it in the state, start it as an empty string.
3. Do not use this direct DOM manipulation, instead set the state as an empty string, never use direct dom manipulation with react components.

onzag (changes requested):
Remove the direct DOM manipulation and the "1" and should be good to go.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/form-element.tsx
avukkonen: Value is sometimes used, and sometimes not. Search component does not need a value, but can be given one -  if I recall correctly. I could make the value a mandatory field and fix the errors that come with it. I think the errors will only be in the organization searches.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/form-element.tsx
```diff
@@ -146,37 +145,59 @@ export class FormWizardActions extends React.Component<FormWizardActionsProps, F
...
   constructor(props: SearchFormElementProps) {
     super(props);
+
+    this.updateSearchField = this.updateSearchField.bind(this);
+    this.clearSearchField = this.clearSearchField.bind(this);
+    this.searchInput = React.createRef();
+
     this.state = {
-      value: this.props.value
+      clearActive: false
```
onzag, line 174: What's the purpose of clearActive? I find no purpose for it, could we remove it?....

onzag (changes requested):

onzag (approved):

PR #5367, opened 2021-02-24
Removed nesting that caused images not to load properly

### LINKED ISSUE #4992: Nested relative sources don't seem to work
If a page has an image within a hyperlink, e.g.

`<a href="hopps5.jpg" onclick="window.open(this.href, '', 'resizable=no,status=no,location=no,toolbar=no,menubar=no,fullscreen=no,scrollbars=no,dependent=no'); return false;">Nelli Nettilukiolaisen HOPS</a>`

...the rendered result is...

`<a href="/workspace/op1-hops-kurssi_51/materials/4_omien_opintojen_suunnittelu/4.6-nelli-nettilukiolaisen-hops/hopps5.jpg"><figure class="image">Nelli Nettilukiolaisen HOPS</figure></a>`

Note how the img src is now empty when it should be the same as the href of a.

onzag:
Fixes #4992 

Nested relative sources are non-standard according to how materials work and as such I decided to normalize these fields.

The way it's done is that.

1. As default every image must be inside a figure element
2. When there's a link on top of a figure element the link is removed.

Let's be careful with side effects but I checked this and everything seems normal after the change.

With the introduction of the zooming element it should be unecessary to have all this extra nesting.

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/base/index.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/base/index.tsx
@@ -124,6 +124,12 @@ function preprocessor($html: any): any{
     }
   });
 
+  $html.find("a figure").each(function(){
+    // removing old style images wrapped in a link
+    // they get processed as link and thus don't work
+    $(this).parent("a").replaceWith(this);
+  });
+
   const $newHTML = $html.map(function() {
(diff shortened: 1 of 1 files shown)


jjlankinen (approved):
Tested and it works just fine. Anchor is removed when it's used to wrap image in CKEditor. Then CKE source is obviously not touched so nothing really breaks as we could revert this rendering change if needed.

What this means is no one can make a link out of images anymore until we fix this in better way.

PR #5395, opened 2021-03-19
5380 dialog wizard page task

### LINKED ISSUE #5380: Dialog wizard page task
When you are on a dialog wizard step, it should be more obvious what you are doing. The task should be stated on top of the step.

Also, there steps of the dialog should be seen. 

Dialog steps should be also divided if the step content pushes the height too high. Most notably, the workspace creation dialog.

avukkonen:
Closes #5380


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/organization/dialogs/edit-workspace.tsx
```diff
@@ -389,13 +392,23 @@ class OrganizationEditWorkspace extends React.Component<OrganizationEditWorkspac
           this.setState({ selectedStudents: this.props.currentWorkspace.studentsSelect.users, studentsLoaded: true });
         }
 
-        return <DialogRow modifiers="edit-workspace">
-          <AutofillSelector modifier="add-students"
-            loader={this.doStudentSearch}
-            placeholder={this.props.i18n.text.get('plugin.organization.workspaces.editWorkspace.search.students.placeholder')}
-            selectedItems={this.state.selectedStudents} searchItems={allItems} onDelete={this.deleteStudent} onSelect={this.selectStudent} />
+        return <DialogRow>
```
jjlankinen, line 395: Case 1 uses `<div>` as a wrapper but here we are using `<DialogRow>` component. This needs to be unified so we use `<div>`, `<Dialogrow>` or `<form>` as a wrapper element.

jjlankinen (changes requested):
Every dialog's edit and create steps should be placed inside `<form>` html element if there is any kind of form elements used in those steps.

Also currently there seem to be 2 `<DialogRow>`components nested and we could either use plain `<div>` element (when no form elements are used in those steps) and/or use `<form>` html element (in steps where we use form elements) as a wrapper.

jjlankinen (approved):

PR #5430, opened 2021-04-12
Sticky search bar gab and content aligning fixed:

### LINKED ISSUE #5421: Sticky search bar has gab with toolbar and content align incorrectly
Näyttökuva 2021-04-07 112947

TeroOtavia:
- application-panel-body.tsx
- Scroll event listener enabled in all device sizes, needed because it
  calculates wheter action bar is sticky or not
- OnScroll listener also calculates everytime offsetElementAgainstTop,
  this fixes gab between action bar and header
- It is possible that this fixes may cause some other bugs, atleast none
  was found when I tested.
- Contains also some very minor code cleaning (const and let variable etc.)

Resolves #5421


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/application-panel.scss
```diff
@@ -122,6 +122,7 @@
     position: relative;
     right: 0;
     top: 0;
+    position: sticky;
```
jjlankinen, line 125: It seems we have 2 position properties. Wonder why Hound did not catch this. Also if `position: sticky` is the one to be used it needs to be placed before `right` property as they need to be placed in alphabetical order.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/application-panel.scss
```diff
@@ -106,8 +106,8 @@
 .application-panel__actions {
   background: $color-default;
   left: 0;
-  position: fixed;
   right: 0;
+  position: fixed;
```
jjlankinen, line 110: `position` property needs to be placed before `right` property. Hound should have spotted this as properties order needs to be alphabetical.

jjlankinen (changes requested):

TeroOtavia:
Issue is closed because currently this up coming fix would break logic with previously fixed scrolling issue in application-panel. They both use same properties to fix different thing different way that would collide together.

TeroOtavia:
Issue is closed because currently this up coming fix would break logic with previously fixed scrolling issue in application-panel. They both use same properties to fix different thing different way that would collide together.

PR #5451, opened 2021-04-20
Drag and drop optimization

### LINKED ISSUE #5050: Reordering material pages in desktop Chrome (83) is wonky
When trying to reorder material pages (and sections) by drag and dropping with mouse is nigh on impossible. Sometimes the dragged content seems to disappear completely, sometimes the drag is left on after the mouse button is released, and during the drag the console is flooded with Uncaught (in promise) crashes. Seems to work a lot more consistently in Firefox, for example.

onzag:
Small optimizations to the drag and drop functionality which should allow it to work in a more efficient way and avoid issues with rendering and it being wonky.

Fixes #5050 to a degree, it is still linked to speed.

Also with devtools drag and drop might work differently.

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
@@ -225,7 +225,7 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
     this.checkAnswers();
   }
   //swapping the fields
-  swapField(triggerChange: boolean, fielda: FieldType, fieldb: FieldType){
+  swapField(executeTriggerChangeFunction: boolean, fielda: FieldType, fieldb: FieldType){
     //if the same then it's pointless
     if (fielda.name === fieldb.name){
       return;
@@ -241,10 +241,10 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
         }
         return f;
       })
-    }, triggerChange ? this.triggerChange : null)
+    }, executeTriggerChangeFunction ? this.triggerChange : null)
   }
   //swapping two counterparts, same as fields
-  swapCounterpart(triggerChange: boolean, fielda: FieldType, fieldb: FieldType){
+  swapCounterpart(executeTriggerChangeFunction: boolean, fielda: FieldType, fieldb: FieldType){
     if (fielda.name === fieldb.name){
       return;
     }
@@ -257,12 +257,12 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
         }
         return f;
       })
-    }, triggerChange ? this.triggerChange : null)
+    }, executeTriggerChangeFunction ? this.triggerChange : null)
   }
   //ok so this is about picking a field, whether it is counterpart or not, and the index it is in
   //I could've found the index it is in, by searching, but I am lazy
   //and I want it in the function
-  pickField(triggerChange: boolean, field: FieldType, isCounterpart: boolean, index: number){
+  pickField(executeTriggerChangeFunction: boolean, field: FieldType, isCounterpart: boolean, index: number){
     //if by the time this function runs there is no selected field
     //then we just set the state that this one is the first selected
     //And we got no business to do
@@ -284,7 +284,7 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
       //then we need to swap two counterparts
       if (this.state.selectedIsCounterpart && isCounterpart){
         //we swap them
-        this.swapCounterpart(triggerChange, this.state.selectedField, field);
+        this.swapCounterpart(executeTriggerChangeFunction, this.state.selectedField, field);
 
         //however we need to figure out for the edited things
         //the opposite of the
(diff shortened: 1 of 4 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
```diff
@@ -225,7 +225,7 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
     this.checkAnswers();
   }
   //swapping the fields
-  swapField(fielda: FieldType, fieldb: FieldType){
+  swapField(triggerChange: boolean, fielda: FieldType, fieldb: FieldType){
```
avukkonen, line 228: fieldA, fieldB?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/draggable.tsx
```diff
@@ -271,6 +281,11 @@ export default class Draggable extends React.Component<DraggableProps, Draggable
     }
   }
   onRootSeletEnd(e: MouseEvent | TouchEvent){
```
avukkonen, line 283: onRootSelectEnd?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
```diff
@@ -241,10 +241,10 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
         }
         return f;
       })
-    }, this.triggerChange)
+    }, triggerChange ? this.triggerChange : null)
   }
   //swapping two counterparts, same as fields
-  swapCounterpart(fielda: FieldType, fieldb: FieldType){
+  swapCounterpart(triggerChange: boolean, fielda: FieldType, fieldb: FieldType){
```
TeroOtavia, line 247: Shorter way to express this is could be:
```
triggerChange && this.triggerChange
```

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
onzag: I don't understand, it's a parameter; why would it be part of "this"?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
onzag: Yes that's what they mean, it would be nicer to name them fieldA and fieldB in camel case but they were originall named like that, I'd have to otherwise rename everywhere they were used.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/draggable.tsx
onzag: Haven't changed that line but the element to drag is called the root in that component, the function triggers on mouse event and touch event; so it's a select event, so the function ended up named on root select end, referring to both mouseend and touchend.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
```diff
@@ -257,12 +257,12 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
         }
         return f;
       })
-    }, this.triggerChange)
+    }, triggerChange ? this.triggerChange : null)
```
TeroOtavia, line 260: Shorter way to express this is could be:
```
triggerChange && this.triggerChange
```

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/draggable.tsx
DerangedPineapple: **onRootSeletEnd** is not the same as suggested **onRootSelectEnd**. Find the typo. Might be better to point the typo in the future if that was the case here.

jjlankinen: onRootSeletEnd is not the same as suggested onRootSelectEnd. Find the typo. Might be better to point the typo in the future if that was the case here.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
avukkonen: I don't see the problem in it, but fair enough. Not a big issue.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
jjlankinen: At least the naming of the variable should be different from the function so it would be easier to read in future.

avukkonen (approved):

PR #5479, opened 2021-04-28
2629 pagination in muikkus announcements panel

### LINKED ISSUE #2629: Pagination in Muikku's announcements panel
At some point there might be quite a few announcements in the front page to users that see all of them so pagination might become useful.

We could limit 10 items per page at first and evaluate further after that.

TeroOtavia:
Announcements panel at frontpage has now pagination:
- added react js pagination component
- contains minor style changes
- contains minor code cleaning

Resolves #2629


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/index/body/announcements-panel.tsx
```diff
@@ -1,69 +1,190 @@
...
+            nextLabel=""
+            breakLabel="..."
+            initialPage={currentPage}
+            marginPagesDisplayed={1}
+            pageCount={pageCount}
+            pageRangeDisplayed={2}
+            onPageChange={this.handlePageChange}
+            ariaLabelBuilder={this.handleAriaLabelBuilder}
+            containerClassName={"pagination"}
+            pageClassName="pager__item pager__item"
```
jjlankinen, line 149: Why `pager__items` is added here twice?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/index/body/announcements-panel.tsx
```diff
@@ -1,69 +1,190 @@
...
+      )
+    })
+
+    /**
+     * renders pagination body as one of announcements list item
+     */
+    const renderPaginationBody = (
+      <div className="item-list__item item-list__item--announcements">
+        <span className="item-list__text-body item-list__text-body--multiline--footer">
+          <ReactPaginate
```
jjlankinen, line 138: There seems to be mixed css classNames/components here. I suggest we use only `pager` components classNames.

Correct structure should be as follows when using our pre-made pager styles:

```
.pager
    .pager__body
        .pager__item pager__item--less icon-arrow-left // If arrow left should be visible
        .pager__item .pager__item--current // If active page
        .pager__item
        .pager__item
        .pager__item pager__item--gap // If there are more pages that gap is needed to display "..."
        .pager__item pager__item--more icon-arrow-right // If arrow right should be visible
```

`.pager__item--first` and `.pager__item--last` might not be applicable with this module. 

ALSO if possible `.pager` could have aria-label added so it follows the same pattern we use in Discussions ie. `aria-label="Tiedotteiden sivutuksen toiminnot/Announcements pagination controls"`

jjlankinen (approved):

jjlankinen (approved):

PR #5502, opened 2021-05-07
404 error from loading avatar is fixed:

### LINKED ISSUE #3845: Don't attempt to load non-existing profile pictures
Discussion areas with lots of messages spams the profile picture url for all senders even if they dont exist. Preload this information from server and avoid doing 404 calls.

TeroOtavia:
- Avatar component is use everywhere where avatar is needed
- contains some minor changes and code cleaning
    - current-thread.tsx
    - discussion-threads.tsx
    - threads.tsx
    - avatar.tsx
    - summary.tsx

Resolves #3845

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/avatar.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/avatar.tsx
@@ -1,20 +1,56 @@
 import * as React from "react";
 import { getUserImageUrl } from "~/util/modifiers";
-import '~/sass/elements/avatar.scss';
+import "~/sass/elements/avatar.scss";
+import { StateType } from "../../reducers/index";
+import { connect, Dispatch } from "react-redux";
+import { i18nType } from "../../reducers/base/i18n";
 
 interface AvatarProps {
-  hasImage: boolean,
-  id: number,
-  firstName: string
+  hasImage: boolean;
+  id: number;
+  firstName: string;
+  userCategory?: number;
+  i18n: i18nType;
+  avatarAriaLabel?: string;
 }
 
-export default function Avatar(props: AvatarProps){
-  let userCategory = props.id > 10 ? props.id % 10 + 1 : props.id;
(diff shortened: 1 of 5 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/avatar.tsx
```diff
@@ -1,20 +1,56 @@
...
-  return props.hasImage ? <object className="avatar-container"
-    data={getUserImageUrl(props.id)}
-    type="image/jpeg">
-     <div className={`avatar avatar--category-${userCategory}`}>{props.firstName[0]}</div>
-  </object> : <div className="avatar-container">
-     <div className={`avatar avatar--category-${userCategory}`}>{props.firstName[0]}</div>
-  </div>
+interface AvatarState {}
+
+class Avatar extends React.Component<AvatarProps, AvatarState> {
```
avukkonen, line 19: This is still a stateless component so it should remain a function?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/avatar.tsx
```diff
@@ -1,20 +1,56 @@
...
 
 interface AvatarProps {
-  hasImage: boolean,
-  id: number,
-  firstName: string
+  hasImage: boolean;
+  id: number;
+  firstName: string;
+  userCategory?: number;
+  i18n: i18nType;
```
avukkonen, line 13: Why does it need i18n? It's not used anywhere.

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/avatar.tsx
TeroOtavia: Removed. Left overs when I tried get i18n locales from redux, but it wasn't necessary.

PR #5551, opened 2021-05-26
fixed drag and drop

onzag:
The drag and drop event triggers twice, the computer is too fast when debugger is not open so react has no time to re-render and the drag function catches twice.

This update stores the original material state into a variable rather than relying on the react state which might be off-sync.


jjlankinen (approved):
Tested locally and works like a charm.

PR #5573, opened 2021-06-02
HTML to react updated and completed mechanism

### LINKED ISSUE #5572: Improved HTML to React
Using the old HTML to React component mechanism, it should be improved to allow proper nesting of all HTML to react properties and elements into the spec.

A spec needs to be written.

onzag:
Fixes #5572


jjlankinen (dismissed):

jjlankinen (approved):
Tested multiple times and works as intended.

PR #5574, opened 2021-06-03
Organization overview for workspaces

### LINKED ISSUE #5568: Organization overview published/unpublished workspaces
child of #5324

saaralepisto:
closes #5568

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/organizationmanagement/rest/OrganizationManagementWorkspaceRESTService.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/organizationmanagement/rest/OrganizationManagementWorkspaceRESTService.java
@@ -1,7 +1,6 @@
 package fi.otavanopisto.muikku.plugins.organizationmanagement.rest;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -222,7 +221,7 @@ public class OrganizationManagementWorkspaceRESTService extends PluginRESTServic
         .setFreeText(searchString)
         .setAccesses(accesses)
         .setAccessUser(loggedUser)
-        .setIncludeUnpublished(true)
+        .setIncludeUnpublished(includeUnpublished)
         .setTemplateRestriction(templateRestriction)
         .setFirstResult(firstResult)
         .setMaxResults(maxResults)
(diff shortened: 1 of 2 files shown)


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/organizationmanagement/rest/OrganizationManagementWorkspaceRESTService.java
```diff
@@ -279,8 +289,11 @@ public Response listWorkspaces(
     } finally {
       schoolDataBridgeSessionController.endSystemSession();
     }
-
-    return Response.ok(workspaces).build();
+    overviewWorkspaces.setPublishedCount(publishedCount);
+    overviewWorkspaces.setUnpublishedCount(unpublishedCount);
+    overviewWorkspaces.setWorkspaces(workspaces);
+    
+    return Response.ok(overviewWorkspaces).build();
```
pkukkon, line 296: This needs its own statistics endpoint just like you did with active/inactive students. You refactored the original search endpoint for workspaces, but you can't just choose to hardcode `includeUnpublished` to `true` as it's a query parameter which the frontend might provide to this endpoint.

You also can't replace the earlier response object (`OrganizationManagerWorkspace`) with an entirely new one (`OrganizationOverviewWorkspaces`) because it breaks all frontend functionality that relies on this endpoint that used to return `OrganizationManagerWorkspace` objects (well, you can if the frontend is adjusted accordingly but there really should be two separate endpoints - one that actually returns workspaces matching the various search criterias and a second one, which returns just the statistics of how many workspaces of the organization are published and unpublished).

This endpoint is used to list organization's workspaces with pagination, maybe free text search, etc. You can't just replace parts of it to also return the total number of published/unpublished courses that the organization has. If you do, but the frontend asks, say, give me just 50 workspaces matching these search terms I gave you, then your statistics will only tell how many of those 50 workspaces were published and unpublished, which is most likely what this feature is not about.

Restore this method to what it was, then create a new one for workspace statistics (once again, as you did with active/inactive students summary). Search from Elastic something like...

```
    searchResult = searchProvider.searchWorkspaces()
        .setOrganizationIdentifiers(organizationIdentifiers) // get this from logged in user, I guess
        .setFirstResult(0)
        .setMaxResults(Integer.MAX_VALUE)
        .search();
```

...and don't do any extra shit when going through the results. Just...

```
if ((Boolean) result.get("published")) {
  publishedCount++;
}
else {
  unpublishedCount++;
}
```

...should suffice.

pkukkon (changes requested):

pkukkon (approved):

PR #5591, opened 2021-06-08
Deleting flag crash fixed:

### LINKED ISSUE #5517: Deletetion of a shared flag crashes
If user tries to delete flag that is shared Muikku throws an error message.

Either prevent shared flag deletion altogether and do not activate the delete button at all in the dialog if the flag is shared OR complete the deletion anyways (this could lead to much confusion with other users that have the shared flag disappeared suddenly).

TeroOtavia:
- Deleting flag crash doesn't crash client
- Deleting requires now deleting all shared links before it can be done
- Added info message if there is selected items to tell that deleting
can't be done
- Persons who have been shared with flags can now delete linking to specific flag

Resolves #5517


petja-laitila:
Merge with devel please, as this is very old already.

avukkonen (approved):

PR #5620, opened 2021-06-15
Build sass with node

aviljakaine:


jjlankinen:
This PR should be redundant as legacy scss files are being removed i separate PR. Also those legacy scss files were not in use anymore after evaluation tools was updated to React.

PR #5623, opened 2021-06-15
File field should now work more reliably:

### LINKED ISSUE #5142: File field does not work reliably
There are 2 issues, one major and one minor.

1. Sometimes when adding files either by dragging them to the file field or by adding them via browser file explorer, the upload throws an error  `Vastauksen tallentamisessa tapahtui virhe (Could not store field value)`. This is fairly easy to reproduce in production and might be related to bad network performance.

2. Quite often also progress bar is not triggered with file upload. This can be reproduced in production easily just by refreshing the page and then trying to upload any file to any file field.

TeroOtavia:
- Progressbar now is correctly shown when adding new files even when given
onchange as prop. It might be little bit a hack solution but it works
- No other problems found when tracking files or adding from file explorer
at the moment

Resolves #5142


petja-laitila:
Still doesn't work. Tested with Chrome and Firefox. After selecting file there is no progress.

onzag (approved):

PR #5651, opened 2021-07-15
prevent joining private chat with oneself

### LINKED ISSUE #5185: Chat: You can start a private chat with yourself
'Nuff said.

onzag:
Fixes #5185

Needs testing, I couldn't make openfire work yet; but the fix is simple enough.

I'd comment if I manage to get openfire to work.


jjlankinen (approved):
Tested and works as it should.

PR #5654, opened 2021-08-06
5653 ghost 404 in audio materials

### LINKED ISSUE #5653: Ghost 404 errors when viewing audio files in materials
there's 2 sets of 404 when there's an audio file in the material page, the audio files are loaded fine in the end, but somehow first they receive a wrong url which leads to error codes in console

avukkonen:
Fixes #5653


jjlankinen (approved):
Tested and works perfectly.

PR #5658, opened 2021-08-11
5646 refator remade evaluation with react

### LINKED ISSUE #5646: Refactor/Remade Evaluation with React
Refactor/Remade Evaluation with using react.

### LINKED ISSUE #5691: Name extension in the workspace dropdown of the new evaluation
If a workspace has once had a name extension but it has since then been removed, the workspace appears in the dropdown as `Workspace name ()`. Assumbaly, when the code determines whether the name extension exists, it checks only undefined and not empty string as well?

### LINKED ISSUE #5692: Incorrect message when going to evaluation in a workspace that has no students
If you have a workspace with no students and go to its evaluation, the page incorrectly states...

"Great! All evaluation requests have been processed!"

...which is not correct as the view is not showing evaluation requests but all students of that workspace (of whom there are none). In this case the message should not be shown, or - preferably - should be something like...

"This workspace has no students"

### LINKED ISSUE #5693: Incorrect dropdown value when going to workspace evaluation of a workspace that isn't yours
The workspace dropdown in evaluation lists all workspaces that you are a teacher in.

However, if you are an admin and go to a workspace in which you are not a teacher, and then launch its evaluation, the dropdown value is left as **All evaluation requests**. This is not correct, as the view is still showing the students of that workspace rather than evaluation requests.

Since the selected workspace wasn't included in the list of the admin's own workspaces, it should probably be dynamically added to the list and then selected, so that the dropdown value would reflect the contents of the view.

### LINKED ISSUE #5703: Workspace's sort order seems to be random in evaluation's dropdown
Evaluation view's left side dropdown that shows workspaces user is part of seems to be listing workspaces randomly and thus resulting different order of workspaces quite often.

We should use same sorting logic here that we are using in course picker to provide consistency across the app.

TeroOtavia:
Resolves #5646 

Bug fixes:
Resolves #5703
Resolves #5691
Resolves #5692
Resolves #5693

Evaluation is now completely refactored with react


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/containers/main-function.tsx
```diff
@@ -106,20 +125,32 @@ export default class MainFunction extends React.Component<MainFunctionProps, {}>
       this.loadCoursePickerData(queryString.parse(window.location.hash.split("?")[1] || "", { arrayFormat: 'bracket' }), true, false);
     } else if (window.location.pathname.includes("/profile")) {
       this.loadProfileData(window.location.hash.replace("#", "").split("?")[0]);
-    }
+    } /* else if (window.location.pathname.includes("/evaluation")) {
```
avukkonen, line 128: Commented code should be removed?

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/evaluation/EvaluationJsPluginMessages_fi.properties
```diff
@@ -1,4 +1,44 @@
 plugin.evaluation.noWorkspaces = Ei ty�tiloja
+plugin.evaluation.notifications.gradesystem.error = Virhe ladattaessa arviointiasteikoita. P\u00E4ivit\u00E4 sivu uudelleen ja yrit\u00E4 uudestaan.
```
avukkonen, line 2: This unicode noise in this file is not a problem?

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/evaluation/EvaluationJsPluginMessages_fi.properties
jjlankinen: No, works just fine. 

However it could be worth while to examine where (and why) those unicodes are coming from. This might be happening when editing locale strings outside of eclipse which should not be a problem.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/evaluation/body/application/evaluation/evaluation-event-content-card.tsx
```diff
@@ -0,0 +1,290 @@
...
+        );
+      default:
+        return;
+    }
+  };
+
+  const parsedDate = moment(date).format("l");
+
+  return (
+    <>
```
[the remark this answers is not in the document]
TeroOtavia: Its shorthand for <React.Fractment></React.Fractment> tag. Multiple elements needs one "container" element as parent, so thats why using fragments. Fragments also doesn't show doesn't add unnecessary nodes to DOM.
[](https://reactjs.org/docs/fragments.html)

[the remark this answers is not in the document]
avukkonen: Ok

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/main-function/evaluation/evaluationActions.ts
```diff
@@ -0,0 +1,1992 @@
...
+
+      dispatch(loadEvaluationAssessmentRequestsFromServer());
+    };
+  };
+
+/**
+ * setEvaluationFilters
+ * @param data
+ */
+let setEvaluationFilters: SetEvaluationFilters = function setEvaluationFilters(
```
onzag, line 1494: Can we not use let anymore if we are not modifying the variables, in the older code you see a lot of let, because a lot was renamed from var just by replacing, and also const was not working for some time for some reason (either misconfiguration or something); we should use const now with newer code, and aim removing all the older lets when we have the chance.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/index.tsx
```diff
@@ -3,199 +3,246 @@
...
-import { bindActionCreators } from 'redux';
-import { UpdateAssignmentStateTriggerType, updateAssignmentState,
-  setWorkspaceMaterialEditorState, SetWorkspaceMaterialEditorStateTriggerType,
-  UpdateWorkspaceMaterialContentNodeTriggerType, updateWorkspaceMaterialContentNode,
-  requestWorkspaceMaterialContentNodeAttachments, RequestWorkspaceMaterialContentNodeAttachmentsTriggerType } from '~/actions/workspaces';
-import { DisplayNotificationTriggerType, displayNotification } from '~/actions/base/notifications';
-
-import '~/sass/elements/rich-text.scss';
-import '~/sass/elements/material-page.scss';
+import * as React from "react";
```
onzag, line 6: Is this just prettier?...

I can't get to find what changed, and I am afraid this could cause merge conflicts with other PRs

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/evaluation/body/application/evaluation/editors/supplementation-editor.tsx
```diff
@@ -0,0 +1,340 @@
...
+    }
+  };
+
+  /**
+   * Component render method
+   * @returns JSX.Element
+   */
+  render() {
+    return (
+      <>
```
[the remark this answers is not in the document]
onzag: React Fragment, used to be unsupported.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/evaluation/dialogs/delete.tsx
```diff
@@ -0,0 +1,228 @@
...
+            onClick={closeDialog}
+          >
+            {this.props.i18n.text.get(
+              "plugin.evaluation.workspaceEvaluationDialog.removeDialog.cancelButton"
+            )}
+          </Button>
+        </div>
+      );
+    };
+    let content = (closeDialog: () => any) => {
```
onzag, line 105: All these kind of let even in the body of render we should aim to change into const.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/evaluation/dialogs/delete.tsx
```diff
@@ -0,0 +1,228 @@
...
+    );
+  }
+}
+
+/**
+ * cleanWorkspaceAndSupplementationDrafts
+ * @param draftId
+ */
+export const cleanWorkspaceAndSupplementationDrafts = (draftId: string) => {
+  if (
```
onzag, line 137: You are retrieving from local storage many times the same value, try to store the in a const; local storage is fast nevertheless, but that's just convention, makes for shorter code and less need for fetching from it all the time.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/containers/workspace.tsx
```diff
@@ -71,43 +111,64 @@ export default class Workspace extends React.Component<WorkspaceProps, Workspace
...
 
     this.state = {
       enrollmentDialogOpen: !props.store.getState().status.loggedIn,
       signupDialogOpen: false,
-    }
+    };
   }
 
+  /**
+   * closeEnrollmentDialog
```
onzag, line 152: What are these for?

The comment is literally the function name.

onzag (changes requested):
Minor changes.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/containers/workspace.tsx
TeroOtavia: Those "comments/documentations" are there now for placeholders only. I haven't look into those function that precisely to document what they are for. I added those just so code is more readable. Previously all functions where one big block of code without clear line breaks. And I think personally that there should be some base for documentation.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/index.tsx
TeroOtavia: Mostly prettier changes. Allthough I changed documentation marking in props because I was annoyed when there where documentation that code editor didn't info me as it could have help me to understand why and what for those props are.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/main-function/evaluation/evaluationActions.ts
TeroOtavia: Fixed. I though there was reason other that var conversion why let where used with some specific components or with redux implementation

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/evaluation/dialogs/delete.tsx
TeroOtavia: Fixed

onzag (approved):

avukkonen (approved):

PR #5674, opened 2021-08-17
Removed misleading field lastResult from SearchResult

### LINKED ISSUE #5374: Misleading field in SearchResult
`SearchResult` object has field `lastResult` that has never been the index of the last result (like `firstResult`) but instead the maximum number of items to return. The entire field could be removed since it serves no purpose; each search result page given to the front-end has first index, an array of result objects, and total hit count, which should be enough for all front-end logic.

saaralepisto:
closes #5374

### DIFF
--- a/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/organizationmanagement/rest/OrganizationUserManagementRESTService.java
+++ b/muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/organizationmanagement/rest/OrganizationUserManagementRESTService.java
@@ -174,7 +174,7 @@ public class OrganizationUserManagementRESTService {
           hasImage));
     }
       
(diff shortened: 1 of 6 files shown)


pkukkon (changes requested):
All react components need to be checked through as well. Some interfaces still refer to both `lastResult` and `maxResults` (and it looks like some calls send `lastResult` as a search parameter to the server even if the server totally ignores it (it would understand `maxResults`, though)).

In other words, the client side is probably equally confusing.

So, if we have 17 items and client shows 10 items per page, the query parameters for the first call would be something like...

`firstResult=0&maxResults=10`

...in which case server responds with...

```
firstResult=0
totalHitCount=17
results=[array of 10 items]
```

...so the second page call from client would be...

`firstResult=10&maxResults=10`

...to which server responds...

```
firstResult=15
totalHitCount=17
results=[array of 2 items]
```

pkukkon (approved):
Backend ok.

aviljakaine (approved):
Seems ok

PR #5714, opened 2021-09-06
Sorting for worklist entries

### LINKED ISSUE #5711: Sorting for worklist entries
Many teacher that fills out worklist entries automatically via evaluation and/or manually via worklist views, might have multiple (read: over 50) entries from a different types and therefore it can get tricky to see possible input errors etc.

Therefore sorting functionality for worklist entries should be added to provide teachers the ability to sort their automatic and manual worklist entries.

Sorting could be alphabetical, by date added and both in reverse. All in all 4 different sorting rules.
* Ascending alphabetical by entry type
* Descending alphabetical by entry type
* Ascending by date created
* Descending by date created

jjlankinen:
Resolves #5711 

Sorting does not get saved and it's section based. Sorting can be done based on entries entryDate and/or their description but separately form each other.

Small icon is added to the right side of the label to indicate when sorting has been done and to which direction (ASC or DESC).


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/profile/body/application/components/work-list-section.tsx
```diff
@@ -0,0 +1,171 @@
...
+      } else {
+        const status = a[actualProperty].localeCompare(b[actualProperty]);
+        // reverse if the direction is wrong
+        return direction === "asc" ? status : -status;
+      }
+    }
+  )
+}
+
+interface IWorkListSectionProps {
```
TeroOtavia, line 29: Consider for code conventions if there are. Prefix "I" is unnecessary and differ from other components. And its not even recommended anymore. So WorkListSectionProps should be fine. And check and remove all prefix from whole pr

TeroOtavia (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/profile/body/application/components/work-list-section.tsx
jjlankinen: Done, don't know why "I" prefix were used but it's removed from every profile view's component (8 files affected in total).

TeroOtavia (approved):

PR #5716, opened 2021-09-07
5530 chat editing room description bugs

### LINKED ISSUE #5186: Chat: Required fields in room creation not enforced
If you try to create a chat room with just name but without description, you only get an error that "creation failed, try again" when the reason of the failure is the lack of description. UI should not allow room creation until both fields have been filled.

### LINKED ISSUE #5530: Chat: editing room description bugs
After one character focus goes to text field (messaging) in background.

TeroOtavia:
- chat group room fields doesn't stole focus from other inputs anymore
- added active group chat room state, which is indicated with box-shadow
- added small validation creation of new chat room and indicators for mandatory fields in form. Cannot create room if mandatory fields are empty

Resolves #5530 
Resolves #5186


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/chat.scss
```diff
@@ -188,6 +189,16 @@
   width: 400px;
 }
 
+.chat__panel--active {
+  box-shadow: 0 0 15px $color-chat-controlbox-box-shadow;
+  transition: box-shadow 0.3s ease-in-out;
+}
+
+.chat__panel--notactive {
```
jjlankinen, line 197: This is unnecessary as it just repeats the default style defined in `.chat__panel` rule already.

jjlankinen (changes requested):

jjlankinen (approved):

PR #5724, opened 2021-09-09
Removed study program selection 

### LINKED ISSUE #5721: Remove study program change dropdown from organization management's user edit dialog
We need to remove/hide study program dropdown from student's edit dialog as the study program change is much more complex than it currently is made to be.

jjlankinen:
Resolves #5721

This removes study program selection from edit student dialog at organization management as the operation the select field was meant to provide is way more complex to be done merely by selecting different study program.

Branch name is misleading as the scope of the issue was changed after branch creation.


pkukkon (commented):
For now, would need to be a hidden field with the original value? Backend will refuse to serve the request without study programme specified (needs refactoring to ignore study programme altogether).

jjlankinen:
This concern is not valid as default study program gets saved to the state and default value is used from there if there has been no change .

avukkonen (approved):

PR #5757, opened 2021-09-27
Optimized script loading

### LINKED ISSUE #5751: Math editor is working very slowly
Currently for some reason math editor is acting very slowly and become quite unusable after a while.

Tested with Linux and latest Firefox and Chrome. Also reported by a user using latest MacOS with latest Safari and Chrome.

### OSTicket
https://helpdesk.muikkuverkko.fi/scp/tickets.php?id=3837

onzag:
Makes scripts not be requested to load twice in certain scenarios, not relying on cache.

Fixes #5751 maybe?...


jjlankinen (approved):
Tested and works as it should.

PR #5825, opened 2021-10-29
5823 taskfield persistent saved indicator

### LINKED ISSUE #5823: Tehtävissä voisi olla pysyvämpi tapa ilmoittaa välitallenneksun onnistumisesta

jjlankinen:
Closes #5823

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/audio-field/index.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/audio-field/index.tsx
@@ -9,6 +9,7 @@ import equals = require("deep-equal");
 import ConfirmRemoveDialog from "./confirm-remove-dialog";
 import Synchronizer from "../base/synchronizer";
 import { AudioPoolComponent } from "~/components/general/audio-pool-component";
+import { FieldStateStatus } from "~/@types/shared";
 
 //so we use the media recorder
 //the media recorder is polyfilled
@@ -61,7 +62,9 @@ interface AudioFieldState {
   //We only set them up in the initial state
   modified: boolean,
   synced: boolean,
-  syncError: string
+  syncError: string,
+
+  fieldSavedState: FieldStateStatus,
 }
 
 //this is the maximum recording time in seconds
@@ -91,6 +94,8 @@ export default class AudioField extends React.Component<AudioFieldProps, AudioFi
       synced: true,
       syncError: null,
 
+      fieldSavedState: null,
+
       //so even
       supportsMediaAPI: this.getSupportsMediaAPI
     }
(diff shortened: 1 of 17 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/connect-field.tsx
```diff
@@ -122,14 +124,22 @@ export default class ConnectField extends React.Component<ConnectFieldProps, Con
...
     this.swapCounterpart = this.swapCounterpart.bind(this);
     this.pickField = this.pickField.bind(this);
     this.cancelPreviousPick = this.cancelPreviousPick.bind(this);
     this.triggerChange = this.triggerChange.bind(this);
+    this.onFieldSavedStateChange = this.onFieldSavedStateChange.bind(this);
+  }
+  onFieldSavedStateChange(savedState: FieldStateStatus){
+    this.setState({
+      fieldSavedState: savedState
+    });
```
TeroOtavia, line 142: Might be nitpicking but for readability consider using spacing between function blocks and even though function that is happening is simple, use js documentation for function even if it is just adding name. It also helps readability by structuring code.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/fields/audio-field/index.tsx
```diff
@@ -424,9 +435,23 @@ export default class AudioField extends React.Component<AudioFieldProps, AudioFi
...
     let ElementDisabledState = this.props.readOnly ? "material-page__taskfield-disabled" : "";
 
+    let fieldSavedStateClass = "";
+    if (this.state.fieldSavedState === "ERROR") {
+      fieldSavedStateClass = "state-ERROR";
+    } else if (this.state.fieldSavedState === "SAVING") {
+      fieldSavedStateClass = "state-SAVING";
+    } else if (this.state.fieldSavedState === "SAVED") {
+      fieldSavedStateClass = "state-SAVED";
+    }
```
TeroOtavia, line 445: These line of codes are used many places. Could consider making one global or at least one level higher function to return that class value as string. Example function could looke like:
```ts
createFieldSavedStateClass = (state: FieldStateStatus) => {
let fieldSavedStateClass = "";
    if (this.state.fieldSavedState === "ERROR") {
      fieldSavedStateClass = "state-ERROR";
    } else if (this.state.fieldSavedState === "SAVING") {
      fieldSavedStateClass = "state-SAVING";
    } else if (this.state.fieldSavedState === "SAVED") {
      fieldSavedStateClass = "state-SAVED";
    }

return fieldSavedStateClass 
}
```
Switch case would work too. And then just:

```ts
let fieldSavedStateClass = createFieldSavedStateClass(this.state.fieldSavedState)
```

TeroOtavia (changes requested):

TeroOtavia (dismissed):

TeroOtavia (dismissed):

TeroOtavia (approved):

PR #5873, opened 2021-11-22
5743 refactor of permissions

### LINKED ISSUE #5743: Refactor permissions
Refactor of the permissions in order to be able to get rid of backing beans.

jjlankinen:
Closes #5743


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/records/body/aside.tsx
```diff
@@ -36,11 +36,11 @@ class Navigation extends React.Component<NavigationProps, NavigationState> {
   isVisible(hash: string) {
     switch (hash) {
       case "hops":
-        return this.props.status.isActiveUser && !this.props.records.studyEndDate && this.props.hops.eligibility && this.props.hops.eligibility.upperSecondarySchoolCurriculum === true;
+        return this.props.status.isActiveUser && this.props.hops.eligibility && this.props.hops.eligibility.upperSecondarySchoolCurriculum === true;
```
TeroOtavia, line 39: Is it necessary to check upperSecondarySchoolCurriculum with `===`  as its type is already boolean, so we can only expect true or false value? So ` && this.props.hops.eligibility.upperSecondarySchoolCurriculum` should be enough. Might be unnecessary nitpicking. Of course the original is more stricter if we can expected value be something else than boolean also.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/reducers/base/status.ts
```diff
@@ -7,6 +7,29 @@
 import $ from '~/lib/jquery';
 import { ActionType } from "~/actions";
 
+export interface WhoAmIType {
```
TeroOtavia, line 10: Would consider if "Type" suffix is necessary use at all for Interfaces. Typescript documentation clearly doesn't use those even with `type alias`. Of course if this has been habit long time, then I understand. Its not that important, but I just wanted to point it out.

TeroOtavia (dismissed):
These are more or less nitpicking. Can be changed if you wish. Otherwise I didn't find anything that would have caught to my eyes.

avukkonen (approved):

aviljakaine (dismissed):

petja-laitila (approved):

PR #5909, opened 2021-12-28
5594-changes:

TeroOtavia:
- added custom feedback element for ckeditor and exercise type assignments
which is shown only if assignments has been reviewed. Also it is hide again
if answer is changed after initial answering

Few pictures how these elements would be rendered in few different cases:
feedback

feedback2

feedback3

feedback4

UPDATE:
feedback-update-FI

feedback-update-EN

Resolves: #5594


ninakurki (approved):
Katselin lähinnä kuvat ja hyvältä näyttää!

avukkonen (approved):

Issue #5920, opened 2021-12-31
Accessing announcements directly causes an error


pkukkon:
If you go to the announcements view (/announcements) directly, loading the view fails, because the url does not have the id of an announcement to be shown. In that case, it should probably show the first announcement in the list and perhaps just an empty view if there are no announcements at all.

PR #5924, opened 2021-12-31
5920 accessing announcements directly causes an error

TeroOtavia:
- announcements with straight url without id will show message that no announcement has been selected or found
- if trying to access announcement with id in url which doesn't exist or user don't have access, showing
corresponding error message
- added two new translation
- added bases for documentation for future use
- some formating

Resolves: #5920


muikku-core-plugins/src/main/resources/fi/otavanopisto/muikku/plugins/announcer/AnnouncerJsPluginMessages_fi.properties
```diff
@@ -4,15 +4,17 @@ plugin.announcements.pageTitle = Tiedotteet
...
+plugin.announcer.errormessage.loadAnnouncements = Tiedotteiden lataus ep�onnistui. Yrit� ladata sivu uudelleen.
+plugin.announcer.errormessage.loadAnnouncement = Tiedotteen lataus ep�onnistui. Yrit� ladata sivu uudelleen.
+plugin.announcer.errormessage.updateAnnouncement = Tiedotteen muokkaus ep�onnistui. Yrit� uudelleen.
+plugin.announcer.errormessage.deleteAnnouncement = Tiedotteen poisto ep�onnistui. Yrit� uudelleen.
+plugin.announcer.errormessage.createAnnouncement = Tiedotteen luominen ep�onnistui. Yrit� uudelleen.
 plugin.announcer.errormessage.createAnnouncement.missing.caption = Tiedotteen otsikko on pakollinen tieto
-plugin.announcer.errormessage.createAnnouncement.missing.content = Tiedotteen sis�lt� on pakollinen tieto
-plugin.announcer.errormessage.createAnnouncement.missing.startDate = Tiedotteen alkamisp�iv�m��r� on pakollinen tieto
-plugin.announcer.errormessage.createAnnouncement.missing.endDate = Tiedotteen loppumisp�iv�m��r� on pakollinen tieto
+plugin.announcer.errormessage.createAnnouncement.missing.content = Tiedotteen sis�lt� on pakollinen tieto
```
jjlankinen, line 15: Encoding problems. Might be because editing those files in vscode and vscode tries to handle them in utf8 encoding or something similar which is wrong.

Locale files should be using ISO-8859-1 encoding.

jjlankinen (changes requested):

jjlankinen (approved):

PR #5932, opened 2022-01-05
Getting workspace education name from Elasticsearch index instead of Pyramus

### LINKED ISSUE #5642: Workspace education type name should be indexed in Elastic
Course picker (for example) fetches education type name of each workspace from Pyramus. It would be a lot more efficient if that information was available in Elastic.

saaralepisto:
Closes #5642


pkukkon (approved):
Requires workspace reindexing after update.

Issue #5305, opened 2021-02-09
Editing and publishing linked and answered material pages


jjlankinen:
If user is editing an linked material page that has task fields which has students answers, it is impossible to publish changes as user ends up in dialog loop where at first confirmation is asked to publish changes to linked materials. Then Muikku asks confirmation to publish changes if it's ok that answers would be lost. Then user ends back to the first dialog.

Reproduce:

1. Create material page that contains task field of some sort.
2. Create copy current workspace.
3. Using another user create an answer to the material page that contains task field.
4. Using admin (or any other user that can edit materials) to edit and publish changes. Remember to focus changes to the task field it self so you can trigger "answers will be lost" confirmation.
5. Observe dialog maddness-loop-a-hoop thingy
6. Frown.

PR #5965, opened 2022-01-21
Dialog fix for materialEditor

TeroOtavia:
Fixing dialog loop for materialEditor when deleting or modifying existing fields in workspace that is linked to other workspaces

Resolves: #5305


jjlankinen (approved):
Tested locally and works just fine

PR #5994, opened 2022-02-04
5992-Evaluation-excercise-side-drawer-panel-fix

TeroOtavia:
Changes:
- excercise panel will not anymore go behind workspace evaluation panel
- side drawer close button can be now disable via props disableClose
- perfecting evaluation jsDoc documentation

Resolves: #5992


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/evaluation.scss
```diff
@@ -539,8 +540,14 @@
     color: $color-assignment;
   }
 
-  .evaluation-modal__evaluate-drawer-close {
+  .button-icon {
```
jjlankinen, line 543: Instead of mixing button styles to evaluation.scss file it's recomended to use --modifiers. 

If we nest css components within other css components then styling will be much harder to read in the future. Also it adds unnecessary specificity which we don't want.

Move .button-icon styles to buttons.scss file and add proper modifier to it. Also remember to add the modifier to the  button's react component also.

I recommend to use something like this:

```
.button-icon--assement-drawer-close {
   ...rules

  &.state-DISABLED {
    ..rules
  }
}
```

Disabled state could be another className alltogether as we normally use .state-DISABLED or something similar to define the styling states for the components.

React component should then render the element with both classNames if supported. 

If IconButton component or Button component cannot handle the extra .state-DISABLED className then we need to either add the support for it OR do the scss like this:

```
.button-icon--assement-drawer-close {
   ...rules
}

.button-icon--assement-drawer-close-disabled {
   ...rules
}
```

This of course targets every `.evaluation-modal__evaluate-drawer-close` instances where said className has been replaced.

jjlankinen (changes requested):

jjlankinen (approved):
Magic Ninja approval even if I added my own changes in here O_o

PR #6009, opened 2022-02-15
Removes JsfResourceRewriteRules

aviljakaine:


jjlankinen (approved):
Tested and everything that was tested was working as usual.

PR #6041, opened 2022-03-04
5645 kurssipoimuriin tieto kurssin pakollisuudesta valinnaisuudesta

TeroOtavia:
Changes:
- added suitability info for course item in coursepicker and for workspace header in workspaces
- solution is party had coded and includes map that contains localizations for different suitability values

Resolves: #5645


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/transcriptofrecords/rest/TranscriptofRecordsRESTService.java
```diff
@@ -646,13 +658,30 @@ public Response listWorkspaces(
     return Response.ok(workspaces).build();
   }
   
+  private Mandatority getMandatority(SchoolDataIdentifier educationSubtypeId) {
```
aviljakaine, line 661: This seems to be the same method in CoursePickerRESTService, TranscriptofRecordsRESTService and WorkspaceRESTService, it should be moved to a place where it can be accessed by all so there's only one instance of the code.

In addition - for every getMandatority -call the educationTypeMapping is loaded from database. This is wasteful if it's done in a loop like it is on line 643. (Same for CoursePickerRestService line 392.)

My suggestion would be: make a getEducationMapping method in a controller (does WorkspaceEntityController make sense? not sure), which reads EducationTypeMapping from database as below and returns it. Then use EducationTypeMapping.getMandatority to ask for the state in the endpoints. Replace all three instances with this approach and the reading of the mapping can be done outside the loop where a loop exists. Also maybe check if there's more places where this generalization can be done.

aviljakaine (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
```diff
@@ -864,6 +874,40 @@ const setCurrentWorkspace: SetCurrentWorkspaceTriggerType =
     };
   };
 
+/**
+ * setAvailableCurriculums
+ */
```
avukkonen, line 879: Missing @returns

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/workspaces/index.ts
```diff
@@ -441,6 +447,10 @@ export interface SetCurrentWorkspaceTriggerType {
   }): AnyActionType;
 }
 
+export interface SetAvailableCurriculumsTriggerType {
```
avukkonen, line 450: Missing JSDoc comment

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/coursepicker/body/application/courses/course.tsx
```diff
@@ -60,6 +63,76 @@ class Course extends React.Component<CourseProps, CourseState> {
     this.toggleExpanded = this.toggleExpanded.bind(this);
   }
 
+  /**
+   * Returns OPS information by curriculum identifier
```
avukkonen, line 67: missing @returns

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/workspace/workspaceHome/header.tsx
```diff
@@ -13,12 +16,24 @@ import {
 
 import "~/sass/elements/hero.scss";
 import "~/sass/elements/meta.scss";
+import { suitabilityMap } from "~/@shared/suitability";
+
+/**
+ * OPSsuitability
+ */
+interface OPSsuitability {
```
avukkonen, line 24: This is already declared in shared. It's not exported from here or ever used.

avukkonen (changes requested):

aviljakaine:
Backend ok

avukkonen (approved):

aviljakaine (approved):

PR #6052, opened 2022-03-14
Labels and flag overflow bug fixed

TeroOtavia:
Changes:
- overlfow issue with flags and labels is now fixed. Added overflow property.
- fixed also small bug with flags in the guider that prevented using arrow keys to navigate between items

Resolves: #5986

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/guider/body/application/toolbar/labels.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/guider/body/application/toolbar/labels.tsx
@@ -32,6 +32,7 @@ import { bindActionCreators } from "redux";
 import { StateType } from "~/reducers";
 import { GuiderType, GuiderStudentType } from "~/reducers/main-function/guider";
 import { ButtonPill } from "~/components/general/button";
+import { AnyActionType } from "~/actions";
 
 /**
  * GuiderToolbarLabelsProps
@@ -94,7 +95,7 @@ class GuiderToolbarLabels extends React.Component<
           items={[
             <div
               key="update-label"
-              className="form-element form-element--new-flag"
+              className="form-element form-element--new-label"
             >
               <input
                 className="form-element__input"
@@ -179,7 +180,10 @@ class GuiderToolbarLabels extends React.Component<
       <Dropdown
         modifier="guider-labels"
         items={[
-          <div key="update-label" className="form-element--new-flag">
+          <div
+            key="update-label"
+            className="form-element form-element--new-label"
+          >
             <input
               className="form-element__input"
               value={this.state.labelFilter}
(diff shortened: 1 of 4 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/guider/body/application/toolbar/labels.tsx
```diff
@@ -176,7 +179,7 @@ class GuiderToolbarLabels extends React.Component<
       <Dropdown
         modifier="guider-labels"
         items={[
-          <div key="" className="form-element">
+          <div key="update-label" className="form-element--new-flag">
```
jjlankinen, line 182: We should name this flag the same as communicator labels as we already use update-label as key attribute's value.

Then there really is no need to add multiple scss rules when we call them labels instead of separately flags. Everything else seems to be named as label internally in the code already.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/guider/body/application/toolbar/labels.tsx
```diff
@@ -92,7 +92,10 @@ class GuiderToolbarLabels extends React.Component<
         <Dropdown
           modifier="guider-labels"
           items={[
-            <div key="update-label" className="form-element">
+            <div
+              key="update-label"
+              className="form-element form-element--new-flag"
```
jjlankinen, line 97: We should name this flag the same as communicator labels as we already use update-label as key attribute's value.

Then there really is no need to add multiple scss rules when we call them labels instead of separately flags. Everything else seems to be named as label internally in the code already.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/form-elements.scss
```diff
@@ -60,6 +60,15 @@
   }
 }
 
+.form-element--new-flag,
```
jjlankinen, line 63: Also remove the new-flag after adjusting the tsx files

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/dropdown.scss
```diff
@@ -143,3 +143,9 @@
     position: relative;
   }
 }
+
+.dropdown--communicator-labels {
+  .dropdown__container {
+    overflow-y: auto;
```
jjlankinen, line 149: Do we need max-height here or is it present somewhere else?

jjlankinen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/sass/elements/dropdown.scss
TeroOtavia: By default dropdown__container has `max-height` property set to 350px

jjlankinen (approved):

PR #6099, opened 2022-05-02
* Guider method findStudent returns field eligibility too

saaralepisto:


muikku-rest/src/main/java/fi/otavanopisto/muikku/rest/model/GuiderStudentRestModel.java
```diff
@@ -0,0 +1,239 @@
+package fi.otavanopisto.muikku.rest.model;
+
+import java.util.Date;
+import java.util.List;
+
+public class GuiderStudentRestModel extends Student {
```
pkukkon, line 6: If you extend a class, you only need to introduce variable/getter/setter for the information you are adding to the base class. Now this class completely overwrites every variable that is already in the base class.

pkukkon (changes requested):

muikku-rest/src/main/java/fi/otavanopisto/muikku/rest/model/GuiderStudentRestModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public GuiderStudentRestModel() {
+  }
+
+  public GuiderStudentRestModel(String id, String firstName, String lastName, String nickName, String studyProgrammeName,
+      String studyProgrammeIdentifier, Boolean hasImage, String nationality, String language, String municipality,
+      String school, String email, Date studyStartDate, Date studyEndDate, Date studyTimeEnd, Date lastLogin,
+      String curriculumIdentifier, boolean updatedByStudent, Long userEntityId, List<StudentFlag> flags,
+      OrganizationRESTModel organization, boolean isUpperSecondarySchoolCurriculum) {
+    super();
```
pkukkon, line 16: None of the base class variables given to this constructor are passed to the constructor of the base class, so they would be all null.

pkukkon (changes requested):

aviljakaine (approved):

pkukkon (approved):

Issue #6100, opened 2022-05-02
iframes should have state to indicate when iframe is active and when its not


TeroOtavia:
Iframes should have invidual state management to indicate when iframe is used and when not. This will resolve problem with mobile devices where iframes biger than device size will prevent scrolling alltogether.

Issue #3072, opened 2017-04-05
Safari in IOS mobile - unable to scroll iframe embeds (PDF)


avukkonen:
User has a problem in IPAD that prevents her from scrolling embedded pdf's in workspace materials (MAA5)

I researched this and found out the problem is common in the mobile safari browser. There might be a way to fix it though, by adding:

-webkit-overflow-scrolling:touch; overflow:auto; 

to the parent container. Easy fix, testing is harder, that's why a separate issue.

avukkonen:
http://helpdesk.muikkuverkko.fi/scp/tickets.php?id=1609

PR #6101, opened 2022-05-02
6100-iframe bug fixed:

TeroOtavia:
Iframe scrolling bug fixed by wrapping it with custom component that
handles outside component click with given onClick props. This fix is only
affecting mobile/pad device sizes

Resolves: #6100
Resolves: #3072

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/static/iframe.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/static/iframe.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import OutsideClickListener from "~/components/general/outside-click-listener";
 import { prepareH5POn } from "~/lib/h5p";
 import { i18nType } from "~/reducers/base/i18n";
 import {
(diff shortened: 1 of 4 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/base/material-loader/static/iframe.tsx
```diff
@@ -147,7 +170,6 @@ export default class Iframe extends React.Component<
             delete iframeProps.width;
             containerStyle = {
               ...containerStyle,
-              paddingTop: "56.25%",
```
jjlankinen, line 150: Removal of this will effectively break 16:9 ratio which can be seen when embedding 16:9 video from youtube.

With `padding-top:  56.25%` the video dimensions are as follows: `width: 658px` and `height: 349px`

When comparing this to devel version, same video that has correct 16:9 aspect ratio has following dimensions: `width: 658px` and  `height: 370px`.

Of course iframe width is always 100% within materials as we process it. Padding-top rule is meant to preserve correct 16:9 aspect ratio of videos when width is artificially adjusted to full width of 100%.

This means that the video will be flattened a bit and therefore makes video fall out of aspect ratio regardless whether the video has black bars or not.

Current solution to add extra element to work as an activating element would mean that it has to inherit it's parent elements styles which means that `.material-page__iframe-wrapper` element could be unnecessary altogether. 

I think solution to this problem is that we either use existing `.material-page__iframe-wrapper` element to function as an state activating element or we transform styles from that element to the `.click-outside-listener ` element and remove the `.material-page__iframe-wrapper` element.

jjlankinen (changes requested):

jjlankinen (approved):

Issue #6112, opened 2022-05-06
frontend redux reducer improvements


TeroOtavia:
Frontend redux implementation is working fine as it is now, but could use little bit improvments from reducer side. These improvments are not currently important and can be made piece by piece if nothing more important matter is going on.

PR #6115, opened 2022-05-11
First iteration reducer improvments. Most of old reducer functions st…

TeroOtavia:
First iteration reducer improvments. Most of old reducer functions still exist but are commented out just in case

Resolves: #6112


avukkonen (changes requested):

avukkonen (dismissed):

jjlankinen (approved):
Has been approved before

Issue #6118, opened 2022-05-13
Editing an existing workspace evaluation with an unknown grade crashes the UI


pkukkon:
Steps to reproduce.

1. In Pyramus, edit some existing course assessment to have a grade from some other scale than 4-10 or S/K (which are the only ones shown in Muikku)
2. View the same evaluation in Muikku. The right side of the view correctly shows the grade.
3. Try to edit the evaluation. The UI crashes.

If an evaluation has a grade that is not normally shown in the evaluation editor, it should probably be added to the grade dropdown on the fly, so that the user can then change it to something "normal" or edit the evaluation so that the grade remains the same.

PR #6122, opened 2022-05-20
6118 editing an existing workspace evaluation with an unknown grade crashes the UI

TeroOtavia:
Bug when using unknown grade system that caused ui to crash is fixed. Unknown grade system is now added to other options if needed

Resolves: #6118


aviljakaine (approved):
Bäkkäri ok

Issue #4928, opened 2020-04-07
Students study line could be added to communicator messages that are viewed by teachers


petja-laitila:
When students send messages to teachers the message should automatically append the students study line information so it's easier for teachers to separate students.

PR #6130, opened 2022-05-25
4928 students study line could be added to communicator messages that are viewed by teachers

TeroOtavia:
Communicator sender name includes now study programme name if sender is student. Changes affect following places:
- messages list view
- message view

Also auto complete component now also shows studyprogramme name for students

Resolves: #4928


aviljakaine (approved):
Backend fine

Issue #6147, opened 2022-06-08
Guider crashes when a study guider does not belong to any user groups


pkukkon:
When a study guider goes to Guider and an attempt is made to fetch their user groups and they don't belong to any, the backend responds with a 403 rather than an empty group list. This response then proceeds to confuse the frontend.

PR #6155, opened 2022-06-09
6147 guider crash

TeroOtavia:
- More specific message if no student is available for guider/teacher
- Filtering headers are also removed if that specific filters are no available

Resolves: #6147


aviljakaine (commented):
Backend ok

jjlankinen (approved):

PR #6157, opened 2022-06-10
Couple material improvments

TeroOtavia:
- too long urls in materials fixed
- new text for excercise hand in buttons
- file upload field new hint text added
- Math toolbar icons are compressed, so it doesn't take so much space

Resolves: #6150 #6151 #6149 #6152


jjlankinen (approved):
Tested manually,. works fine after few changes.

PR #6158, opened 2022-06-13
Whatsapp functionality and extra info field for profile

TeroOtavia:
- Whatsapp functionality for teachers. In profile it can be toggled on/off whether students can use that to contact teachers from workspace view
- Extra info field for teachers in profile view. Here they can added any extra info they need to like links etc. This info is then shown in teachers item at workspaces view.

Resolves: #6132


jjlankinen (approved):
Tested manually and works fine after few minor styling changes

Issue #6165, opened 2022-06-17
Hops calculation and visibility bug


TeroOtavia:
There is small calculation bug in compulsory hops and visibility issue with student that should not see that hops at all

PR #6167, opened 2022-06-17
Hops visibility and mandatory calculation bug fixed

TeroOtavia:
- visibility bug fixed
- mandatory course calculation fixed

Resolves: #6165


jjlankinen (approved):
Tested manually and confirmed everything works

PR #6175, opened 2022-07-28
5093 linkit tieto uuteen valilehteen avautumisesta

TeroOtavia:
Added accessibility icons to indicate if links opens to another tab or window.

Changes made to:
frontpage
profile icon list

Resolves: #5093


TeroOtavia:
Commit is missing translations for possible hidden screen reader text elements

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/button.tsx
```diff
@@ -87,10 +97,18 @@ export default class Button extends React.Component<ButtonProps, ButtonState> {
           .map((s) => `button--${s}`)
           .join(" ")}`}
       >
-        {this.props.icon && (
-          <span className={`button__icon icon-${this.props.icon}`}></span>
+        {this.props.icon && this.props.iconPosition === "left" && (
```
avukkonen, line 100: This could be done in css with flexbox order with just a button__icon--left or button__icon--right classes.

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/button.tsx
jjlankinen: There currently are `button__icon--left` and `button__icon--right` classes and they provide correct margin's for the said button based on the iconPosition value.

This means that if iconPosition has value of `left` it has different margins that `right` value. It has nothing to do with flexbox.

avukkonen: I was thinking about something like this. 

```
button__icon--left {
  margin-right: 10px;
  order: -1;
}

button__icon--right {
  margin-right: 10px;
  order: 1;
}
```

avukkonen (approved):

Issue #6198, opened 2022-08-30
Reading helper functionalities and reading ruler


TeroOtavia:
Add helper functionalities to muikku, both enviromental and workspace level. Issue consist parts:

1. Create helper functionalities component that contains every possible helper function for users to use. The most convienient way to handle these functions is probably redux or something else maybe.

2. Create and add reading ruler. It should have following functions: changes height of ruler, change background color of ruler elements, invert ruler colors, pin ruler to place and for last click to close ruler.

3. Add reading ruler to helper functionalities, with possible couple default options and custom option.

PR #6209, opened 2022-09-05
6198 reading helper functionalities and reading ruler

TeroOtavia:
- added easy-to-use functions as side drawer. Place might changes later on...
- added reading ruler components
- Reading ruler has preset support, localstorage support, change color support,
change width support, invert colours support, overlayclick support, pin to place support,
follow cursor support and for mobile drag handle support.
- Ruler component is separated to actual base component, portal and controller components.
This is so developer tools will recognize reading ruler component and its states
- Includes easy-to-use reducer to handle possible future functionalities
- Easy-to-use added to index page, main-function and workspace. All these shares
ruler settings via localstorage

Resolves: #6198


jjlankinen (approved):
Tested and works fine. Cannot really brake anything.

PR #6256, opened 2022-10-03
6246 workspace member material restriction

TeroOtavia:
Added new view restriction option for material editor -> workspace members only.

Resolves: #6246


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/workspace/workspaceHelp/help/index.tsx
```diff
@@ -31,12 +32,15 @@ import {
...
 } from "~/actions/workspaces";
 import { Redirect } from "react-router-dom";
+import { StatusType } from "~/reducers/base/status";
 
 /**
  * HelpMaterialsProps
  */
 interface HelpMaterialsProps {
   i18n: i18nType;
+  isStudent: boolean;
```
avukkonen, line 42: does this need to be here, since it's already under the status prop?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/workspace/workspaceHelp/help/index.tsx
```diff
@@ -732,11 +763,13 @@ class Help extends React.Component<HelpMaterialsProps, HelpMaterialsState> {
 function mapStateToProps(state: StateType) {
   return {
     i18n: state.i18n,
+    status: state.status,
     workspace: state.workspaces.currentWorkspace,
     materials: state.workspaces.currentHelp,
     activeNodeId: state.workspaces.currentMaterialsActiveNodeId,
     workspaceEditMode: state.workspaces.editMode,
     isLoggedIn: state.status.loggedIn,
+    isStudent: state.status.loggedIn && state.status.isStudent,
```
avukkonen, line 772: what's the point of checking if the user is logged in? 

isStudent is false if you are not logged in
As far as I see it, you always check if the user is logged in anyway every time you check if user is a student

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/workspace/workspaceMaterials/materials/index.tsx
```diff
@@ -773,12 +805,14 @@ class WorkspaceMaterials extends React.Component<
...
   return {
     i18n: state.i18n,
+    status: state.status,
     workspace: state.workspaces.currentWorkspace,
     materials: state.workspaces.currentMaterials,
     materialReplies: state.workspaces.currentMaterialsReplies,
     activeNodeId: state.workspaces.currentMaterialsActiveNodeId,
     workspaceEditMode: state.workspaces.editMode,
     isLoggedIn: state.status.loggedIn,
+    isStudent: state.status.loggedIn && state.status.isStudent,
```
avukkonen, line 815: You need the loggedIn here?

Also, you have the same infromations 2 times in the props and by doing that, you save writing "this.props.status.isStudent" instead of "this.props.isStudent". The difference is not huge, but you end up mapping more stuff to props. 

I could not find a single case about what's the best practice here, so I guess this is fine. Just wondering if having the same properties many times in the props is actually a good practice.

avukkonen (changes requested):

avukkonen (approved):

pkukkon (approved):

PR #6265, opened 2022-10-06
6203 lukion hops lomake nakymaan ohjaamoon

TeroOtavia:
Reworked guider hops little bit so data is shown now correctly. Previously uppersecondary hops didn't show same data what student sees in records hops view. Also all kind of vops and hops-enabled functionality is removed from both frontend and backend.

Resolves: #6203


aviljakaine (dismissed):
Backend ok

avukkonen (changes requested):
There are many of these ESLint warnings that are fixable with a simple find / replace. In a sense they are outside of the scope though.

avukkonen (approved):

PR #6291, opened 2022-10-19
6247 journal as a new assignment type

TeroOtavia:
Material has now new page type (JOURNAL). With that and with new Ckeditor field (JournalField) users can create "mandatory" journal assignments to material. When students writes something to that field its automatically creates new journal entry to journal view also.

Resolves: #6247


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/material/QueryJournalFieldController.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public QueryJournalField createQueryJournalField(Material material, String name) {
+    return queryJournalFieldDAO.create(material, name);
+  }
+
+  public QueryJournalField findQueryTextFieldbyId(Long id) {
+    return queryJournalFieldDAO.findById(id);
+  }
+
+  public QueryJournalField findQueryTextFieldByMaterialAndName(Material material, String name) {
```
pkukkon, line 31: findQueryJournalFieldByMaterialAndName

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/WorkspaceMaterialFieldAnswerController.java
```diff
@@ -116,6 +116,21 @@ public WorkspaceMaterialTextFieldAnswer updateWorkspaceMaterialTextFieldAnswerVa
     return workspaceMaterialTextFieldAnswerDAO.updateValue(workspaceMaterialTextFieldAnswer, value);
   }
   
+  /* JournalField */
```
pkukkon, line 119: Are these three methods used at all? Doesn't WorkspaceJournalFieldIOHandler create, update, and find field data via WorkspaceJournalController?

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/fieldio/WorkspaceJournalFieldIOHandler.java
```diff
@@ -0,0 +1,73 @@
...
+        UserEntity userEntity = userEntityController.findUserEntityById(reply.getUserEntityId());
+        
+          Long workspaceEntityId = workspaceMaterialController.getWorkspaceEntityId(field.getWorkspaceMaterial());
+          WorkspaceEntity workspaceEntity = workspaceEntityController.findWorkspaceEntityById(workspaceEntityId);
+          
+          if (workspaceEntity != null) { 
+          journalEntry = workspaceJournalController.createJournalEntry(workspaceEntity, userEntity, value, identifier, field.getWorkspaceMaterial().getTitle());
+        }
+      } else {
+        journalEntry = workspaceJournalController.updateJournalEntry(journalEntry, field.getWorkspaceMaterial().getTitle(), value);
```
pkukkon, line 47: For title, maybe use journalEntry.getTitle() instead? If the entry exists and the user edited its title in the journal, that title wouldn't then revert back to page title if the user edits the entry in materials.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/workspace/fieldio/WorkspaceJournalFieldIOHandler.java
```diff
@@ -0,0 +1,73 @@
...
+      }
+    } else {
```
pkukkon, line 49: Is this if..else block needed? Even if the value would be blank, shouldn't it still lead to a journal entry being either created or updated?

pkukkon (changes requested):
Minor backend cleanup work. CKEditor plugin seems OK.

updates/1.124/221006_journalEntry.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<update xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.ofw.fi/xml/2011/java-xmldb-updater/UpdaterSchema.xsd">
+
+  <sql>alter table workspaceJournalEntry add column materialFieldReplyIdentifier varchar(255) after created;</sql>
```
aviljakaine, line 4: Could capitalize the W in table name (another case on the next line), some database instances are case sensitive.

avukkonen (changes requested):

pkukkon (approved):
Backend OK.

avukkonen (approved):

PR #6336, opened 2022-11-11
changes locales for the wall tasks

### LINKED ISSUE #6334: Seinämän tyhjän tehtävälistan lokalisointi
Parempi lokalisointi

avukkonen:
Closes #6334


jjlankinen (approved):
Empty class element normally needs <span> or some other html element to be placed inside of it but in this case it's ok to brake the rule for now. Maybe later introduce different styling for different places as currently normal way would introduce too big of an font for the empty text to be used in index page.

Issue #6340, opened 2022-11-14
Change frontend language picker to use rest


TeroOtavia:
Change language picker to use rest

Issue #6269, opened 2022-10-06
Refactor language selection to rest


aviljakaine:
Refactor language selection away from base.xhtml and give it a rest endpoint instead.

PR #6352, opened 2022-11-16
6340 change frontend language picker to use rest

TeroOtavia:
Changes:
- changing localization is done by using mApi endpoint and any related unnecessesary jsf files are removed
- couple meta tags are also moved to whoami endpoint as part of moving towards separating frontend and backend

Resolves: #6340 
Resolves: #6269


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/me/MeRESTService.java
```diff
@@ -72,10 +72,28 @@ public class MeRESTService {
...
+   * {
+   *    lang: "en"
+   * }
+   */
+  @GET
+  @Path("/locale")
+  @RESTPermit (handling = Handling.UNSECURED)
+  public Response getLocale() {
+    Locale locale = localSessionController.getLocale();
+    String lang = (locale == null || locale.getLanguage() == null) ? "FI" : locale.getLanguage().toUpperCase();
```
pkukkon, line 92: Shouldn't this be lowercase to indicate language? After all, only country names are in uppercase (e.g. `en-GB` or `en-US`)

pkukkon (changes requested):

pkukkon (dismissed):
Backend OK.

avukkonen (approved):

PR #6489, opened 2023-02-14
6488-Lang-attribute-after-review-and-testing-fixes

TeroOtavia:
- material editor lang attribute fixed. Added more language options
- affects to help, material, workspace home and management views
- language code locale changed to be just "Language" instead of "Language code"
- language option locales are missing but are added after i18next is implemented

Resolves: #6488


saaralepisto (approved):
backend ok

jjlankinen (approved):

PR #6595, opened 2023-05-31
Materials are disabled if there is any type pending request ongoing

TeroOtavia:
Materials are disabled if there is any type pending request ongoing

Resolves: #4596


jjlankinen (approved):
Tested manually to be in working order

Issue #6609, opened 2023-06-07
Language options uses wrong locale code for some options


TeroOtavia:
Some of the Material editor lang attribute select options uses wrong locale code. This must be fixed by creating script to fix values in database and correct options in frontend.

PR #6610, opened 2023-06-07
6609-Language codes are fixed

TeroOtavia:
Language codes are fixed.

Resolves: #6609


pkukkon (approved):

jjlankinen (approved):
Locale codes are correct

Issue #6637, opened 2023-08-23
Announcement openapi-generator rework


TeroOtavia:
Replace old mApi announcement related api calls with newly generated typescript-fetch calls.

PR #6642, opened 2023-08-28
6637 announcement openapi generator rework

TeroOtavia:
Announcements side of muikku is converted to use new MApi created by using openapi-generator and OpenApi swagger.yaml document. No additional logic refactoring done, just api fetches changed

Resolves: #6637

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
@@ -102,9 +102,6 @@ export async function loadAnnouncementsHelper(
   try {
     const announcements = await announcerApi.getAnnouncements(params);
 
-    /* const announcements = <Announcement[]>(
-      await promisify(mApi().announcer.announcements.read(params), "callback")()
-    ); */
     announcements.forEach((a) =>
       a.userGroupEntityIds.forEach((id) => dispatch(loadUserGroupIndex(id)))
     );
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
@@ -4,12 +4,11 @@ import {
   AnnouncementsStatePatch,
   AnnouncerNavigationItemType,
 } from "~/reducers/announcements";
-import promisify from "~/util/promisify";
-import mApi, { MApiError } from "~/lib/mApi";
+import { MApiError } from "~/lib/mApi";
 import notificationActions from "~/actions/base/notifications";
 import { StateType } from "~/reducers";
 import { loadUserGroupIndex } from "~/actions/user-index";
-import { Announcement, GetAnnouncementsRequest } from "~/generated/client";
+import { GetAnnouncementsRequest } from "~/generated/client";
 import { Dispatch } from "react-redux";
 import MApi from "~/api/api";
 
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
@@ -1,18 +1,17 @@
 import { AnyActionType } from "~/actions";
 import {
-  AnnouncementsType,
   AnnouncementsStateType,
-  AnnouncementListType,
-  AnnouncementsPatchType,
-  AnnouncerNavigationItemListType,
+  AnnouncementsStatePatch,
   AnnouncerNavigationItemType,
 } from "~/reducers/announcements";
-import { StatusType } from "~/reducers/base/status";
 import promisify from "~/util/promisify";
 import mApi, { MApiError } from "~/lib/mApi";
 import notificationActions from "~/actions/base/notifications";
 import { StateType } from "~/reducers";
 import { loadUserGroupIndex } from "~/actions/user-index";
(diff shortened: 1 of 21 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/api/api.ts
```diff
@@ -0,0 +1,275 @@
...
+  WorkspaceDiscussionApi,
+  WorkspaceNotesApi,
+} from "../generated/client";
+
+const isDevelopment = process.env.NODE_ENV !== "production";
+
+const configuration = new Configuration({
+  basePath: isDevelopment
+    ? window.location.origin
+    : "https://otavanopisto.muikkuverkko.fi",
```
aviljakaine, line 35: Hard-coded domain is not going to work with installations on different domains (duh).

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/api/api.ts
```diff
@@ -0,0 +1,275 @@
...
+  WorkspaceNotesApi,
+} from "../generated/client";
+
+/* const isDevelopment = process.env.NODE_ENV !== "production"; */
+
+const configuration = new Configuration({
+  basePath: window.location.origin,
+});
+
+console.log("API base path: " + configuration.basePath);
```
aviljakaine, line 36: Is the logging necessary?

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/package.json
```diff
@@ -5,12 +5,12 @@
...
   "scripts": {
     "build": "webpack",
+    "build-full": "npm run build-client && npm run build",
     "test": "echo \"Error: no test specified\" && exit 1",
     "eslint": "eslint . --ext .ts,.tsx",
     "stylelint": "stylelint \"**/*.scss\"",
     "prettier-format": "prettier --config .prettierrc.json \"**/*.{ts,tsx}\" --write",
-    "build-client": "./node_modules/@openapitools/openapi-generator-cli/bin/openapi-generator generate -i ../../swagger.yaml -o generated/client -c generator-config.json -g typescript-fetch",
-    "build-client-windows": "openapi-generator-cli generate -i ../../swagger.yaml -o generated/client -c generator-config.json -g typescript-fetch"
+    "build-client": "openapi-generator-cli generate -i ../../swagger.yaml -o generated/client -c generator-config.json -g typescript-fetch"
```
jjlankinen, line 13: `npm run build-full` calls `build-client` script which seems to output `generated/client` folder even with full maven build as it is called in `pom.xml` right after `npm ci` is completed. 

Is this the correct way to build something to production as it seems that generated folder is included in the prod. build even if it was meant to be used in development only.

Previously we used `NODE_:ENV` variable to control how webpack made the build.

TeroOtavia: You need that client to build frontend because frontend has references to generated client code "aka" interfaces and fetch methods. Without those frontend is not completed and thus cannot be build properly. 

What is not necessary is that generated/client is present in github, because we have swagger.yaml that holds every data to generate client when ever needed. This means that only when you are developing or building production you need to generated that client.

Why is pom.xml changed to use `npm run build-full`, because so that `mvn clean compile` which also builds frontend needs to have client present for the reasons previously mentioned. This should also simplify how backend developers can get frontend running without running multiple different commands separately on the frontend side. For frontend developers you can still run `mvn clean compile` once and from that point just run `npm run build` whenever you have someother frontend changes.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
```diff
@@ -100,17 +97,21 @@ export async function loadAnnouncementsHelper(
       break;
   }
 
+  const announcerApi = MApi.getAnnouncerApi();
+
   try {
-    const announcements: AnnouncementListType = <AnnouncementListType>(
+    const announcements = await announcerApi.getAnnouncements(params);
+
+    /* const announcements = <Announcement[]>(
```
avukkonen: Commented code

petja-laitila (approved):
Reviewed build process (pom and package.json).

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/containers/main-function.tsx
```diff
@@ -131,6 +130,7 @@ interface MainFunctionProps {
   websocket: Websocket;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 (window as any).USES_HISTORY_API = true;
```
avukkonen, line 134: You can get rid of the any with extending  the window Ambient type

```
declare global {
    interface Window { 
         USES_HISTORY_API: true
    }
}
window.USES_HISTORY_API = true
```

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/containers/main-function.tsx
TeroOtavia: Is this really necessary to fix in this branch? Outside side of whole branch scope anyways. Can be done later with separate feature branch

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/announcements/helpers.ts
TeroOtavia: Already fixed

avukkonen (approved):

aviljakaine (approved):

Issue #6643, opened 2023-08-29
Communicator openapi-generator rework


TeroOtavia:
Replace old mApi communicator related api calls with newly generated typescript-fetch calls.

PR #6657, opened 2023-09-01
6643 communicator openapi generator rework

TeroOtavia:
Communicator api refactor done. Added also MApiError type guards to api folder. 

Resolves: #6643

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/communicator/body/application/message-view/message.tsx
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/communicator/body/application/message-view/message.tsx
@@ -1,13 +1,9 @@
 import * as React from "react";
-import {
-  MessageType,
-  MessageThreadLabelListType,
-} from "~/reducers/main-function/messages";
 import Link from "~/components/general/link";
 import { StateType } from "~/reducers";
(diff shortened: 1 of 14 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/communicator/body/application/message-view/message.tsx
```diff
@@ -277,7 +280,7 @@ class Message extends React.Component<MessageProps, MessageState> {
       this.props.message.userGroupRecipients.map(
         (ug): ContactRecipientType => ({
           type: "usergroup",
-          value: ug,
+          value: ug as any,
```
avukkonen, line 283: Can the any be avoided?

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/communicator/body/application/message-view/message.tsx
TeroOtavia: Not specifically in this branch thats why type any. It will be resolved following branches, when more usergroup related paths and interfaces are replaced with generated api.

avukkonen (approved):

PR #6670, opened 2023-09-11
Logged in status fix which prevented staff and announcmenets loading …

TeroOtavia:
Logged in status fix which prevented staff and announcmenets loading occasionally. This happened especially with slow interned speed connections

Resolves: #6666 #6665


jjlankinen (approved):
Tested locally via throttling and issue is fixed

Issue #6678, opened 2023-09-15
Records openapi-generator rework 


TeroOtavia:
Replace old mApi records related api calls with newly generated typescript-fetch calls.

PR #6679, opened 2023-09-15
Records api path refactor done

TeroOtavia:
Records api path refactor done

Resolves: #6678


avukkonen (approved):

muikku-core-plugins/src/main/resources/META-INF/resources/swagger.yaml
```diff
@@ -12770,7 +12771,7 @@ components:
 
     RecordWorkspaceActivityInfo:
       type: object
-      description: RecordsWorkspaceActivityInfo object
+      description: Record workspace activity info with line and category, default line and activities
       required:
         - lineName
         - lineCategory
```
avukkonen, line 12777: These are not "lines", these are studyPrograms

avukkonen (changes requested):

muikku-core-plugins/src/main/resources/META-INF/resources/swagger.yaml
TeroOtavia: Rest speaks about lines.

muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/actions/main-function/records/index.ts
```diff
@@ -8,13 +8,13 @@ import {
   TranscriptOfRecordLocationType,
   CurrentStudentUserAndWorkspaceStatusType,
   CurrentRecordType,
-  RecordWorkspaceActivityInfo,
   RecordWorkspaceActivityByLine,
```
avukkonen, line 11: Out of scope, but this needs to be addressed somewhere. These are not lines, they are studyPrograms.

muikku-core-plugins/src/main/resources/META-INF/resources/swagger.yaml
avukkonen: I'll make an issue about this. Carry on.

avukkonen (approved):

PR #6688, opened 2023-09-19
6672 pedagogisen tuen suunnitelma lukutapahtamien lokitus

TeroOtavia:
Pedagogy form now logs histry events with following types: "VIEW" and "EDIT". VIEW logs are fired whenever pedagogy form is opened and EDIT when form is saved. History events have simple filter to address these changes when rendering history event list.

Resolves: #6672


updates/1.139/190923_pedagogy_form_history_type.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<update xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.ofw.fi/xml/2011/java-xmldb-updater/UpdaterSchema.xsd">
+
+  <sql>alter table pedagogyformhistory add column type varchar(255) not null;</sql>
+  <sql>update pedagogyformhistory set type = "EDIT" where type = null;</sql>
```
pkukkon, line 5: This won't update anything. Because type is a not null column, values for previous rows will be empty strings. Can simply be...

`update PedagogyFormHistory set type='EDIT';`

Also capitalize the table name in the previous statement (might be that Linux/MariaDB is stricter about this).

pkukkon (changes requested):

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/pedagogy/PedagogyController.java
```diff
@@ -253,7 +253,22 @@ public PedagogyForm findFormByStudentIdentifier(String studentIdentifier) {
   }
   
   public void createViewHistory(PedagogyForm form, Long modifierId) {
+    List<PedagogyFormHistory> historyList = pedagogyFormHistoryDAO.listByFormAndCreator(form, modifierId);
```
pkukkon, line 256: Might as well list by form, creator, and type (VIEW)

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/pedagogy/PedagogyController.java
```diff
@@ -253,7 +253,22 @@ public PedagogyForm findFormByStudentIdentifier(String studentIdentifier) {
...
+    
+    if (historyList != null) {
```
pkukkon, line 258: Technically not needed, as DAO methods are guaranteed to always return an empty list.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/pedagogy/PedagogyController.java
```diff
@@ -253,7 +253,22 @@ public PedagogyForm findFormByStudentIdentifier(String studentIdentifier) {
...
+      PedagogyFormHistory latestHistoryItem = historyList.get(0);
```
pkukkon, line 259: This will crash if the DAO method returns an empty list.

pkukkon (changes requested):

pkukkon (approved):
Backend OK.

pkukkon (approved):

PR #6770, opened 2023-10-13
Added maven-clean-plugin to remove the compiled frontend in clean phase

aviljakaine:
This cleans the compiled frontend files (generated and dist) in maven clean phase. As the frontend compilation doesn't seem to reuse the files in any way (like partial recompilation etc), it could be a useful addition to ensure that the stage is clean for a full build.

Are there any unintended consequences that come to mind?

### DIFF
--- a/muikku-core-plugins/pom.xml
+++ b/muikku-core-plugins/pom.xml
@@ -259,6 +259,29 @@
   <build>
     <plugins>
       <plugin>
+        <artifactId>maven-clean-plugin</artifactId>
+        <version>3.3.1</version>
+        <configuration>
+          <filesets>
+            <fileset>
+              <directory>src/main/resources/META-INF/resources/scripts/dist</directory>
+              <includes>
+                <include>**/*</include>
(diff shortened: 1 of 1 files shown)


muikku-core-plugins/pom.xml
```diff
@@ -258,6 +258,29 @@
...
+              <directory>src/main/resources/META-INF/resources/scripts/src/generated</directory>
+              <includes>
+                <include>**/*</include>
+              </includes>
+              <followSymlinks>false</followSymlinks>
+            </fileset>
+          </filesets>
+        </configuration>
+      </plugin>
+
```
petja-laitila, line 283: At the moment deploy is done with clean. Not sure how this affects that. Why are we also cleanind dist?

petja-laitila: Maybe we should remove clean from deploy phase? I don't really see the need for it.

pkukkon (approved):

PR #6773, opened 2023-10-16
Annotation-related warning fixes

aviljakaine:
Fixed a bunch of annotation-related warnings that were present in console

* `GuidanceRequest` / `WorkspaceGuidanceRequest` had inheritance backwards, inheritance strategy should be specified in the superclass and it's completely ignored in the subclass
* `@PrimaryKeyJoinColumn` was specified in classes that didn't have JOINED or any inheritance strategy what so ever. This should be used only in cases with JOINED inheritance strategy
* Cache annotations are ignored (as per warn in console) for subclasses so they were removed. In these cases the superclass had the same annotations which are still present


pkukkon (approved):
Seems legit.

Issue #6849, opened 2023-10-27
Dialog component with focus trap and other accessiblity functionality


TeroOtavia:
Add focus trap functionality and other needed accessibility functionalities to dialog like aria-label and roles etc.

child of #6835

PR #6853, opened 2023-10-27
6849 dialog component with focus trap and other accessiblity functionality

TeroOtavia:
Changes:
- Focus trap added to Dialog component
- close link replaced with Iconbutton
- some aria label refactoring (checked using axe chrome extension)

Resolves: #6849


muikku-core-plugins/src/main/resources/META-INF/resources/scripts/src/components/general/dialog.tsx
```diff
@@ -136,71 +138,77 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
...
-                .join(" ")} ${this.state.visible ? "dialog--visible" : ""}`}
-              onClick={
-                closeOnOverlayClick
-                  ? this.onOverlayClick.bind(this, closePortal)
-                  : null
-              }
-            >
-              {/* Execution container is missing from here */}
-              <section
+            <FocusTrap active={this.state.visible}>
```
jjlankinen, line 141: FocusTrap prevents second dialog from opening which is used when creating a new purchase for study time period increase.

Problem might be somewhat related to UX/UI design also as I think opening dialog within a dialog is a bit questionable and always harder to handle accessibility wised. Currently there's nothing to be done about it and FocusTrap needs to be reworked to support this functionality but in the future I think we should avoid this kind of thing.

jjlankinen (changes requested):
Problems that UI tests caught will not be fixed until FocusTrap related issues are solved.

avukkonen (approved):

jjlankinen (approved):

PR #6857, opened 2023-10-31
6477 opintopisteet ohjaamoon

### LINKED ISSUE #6477: Opinnot ja Ohjaamo: Opintopisteet näkyviin
Suoritusrekisterissä pitää nähdä opintopisteet
--> pakolliset, kaikki yhteensä

saaralepisto:
closes #6477


muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/transcriptofrecords/rest/TranscriptofRecordsRESTService.java
```diff
@@ -167,9 +178,105 @@ public Response listWorkspaceActivities(
...
+                  // Transferred courses doesn't have ids or identifiers so that's why these need to get separately
+                  if (activity.getId() == null && assessmentState.getState() == WorkspaceAssessmentState.TRANSFERRED) {
+                    Mandatority mandatority = activity.getMandatority();
+                    if (mandatority != null && mandatority == Mandatority.MANDATORY) {
+                      mandatoryCourseCredits = Integer.sum(units, mandatoryCourseCredits);
+                   }
+                  }
+                  
+                  // Search for finding out course mandaority
+                  SearchProvider searchProvider = getProvider("elastic-search");
```
aviljakaine, line 215: Move this outside all loops, it always resolves to same result and is not depending on any loop variables.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/transcriptofrecords/rest/TranscriptofRecordsRESTService.java
```diff
@@ -167,9 +178,105 @@ public Response listWorkspaceActivities(
...
+    activityInfo.setMandatoryCourseCredits(mandatoryCourseCredits);
+    activityInfo.setShowCredits(false);
+    
+    User user = userController.findUserByDataSourceAndIdentifier(studentIdentifier.getDataSource(), studentIdentifier.getIdentifier());
+    
+    // Find student's curriculum to tell whether the score will be shown to the user
+    
+    String curriculumName = guiderController.getCurriculumName(user.getCurriculumIdentifier());
+    
+    if (curriculumName != null && curriculumName.equals("OPS 2021") && (activityInfo.getLineCategory() != null && activityInfo.getLineCategory().equals("Lukio"))) {
```
aviljakaine, line 264: Again, I assume this could be used as the major flag if any credit counts need to be resolved at all.

muikku-core-plugins/src/main/java/fi/otavanopisto/muikku/plugins/guider/GuiderRESTService.java
```diff
@@ -690,18 +705,19 @@ public Response listWorkspaces(
...
-
-    User user = userController.findUserByDataSourceAndIdentifier(userIdentifier.getDataSource(), userIdentifier.getIdentifier());
+    torWorkspaceWithCourseCredits.setShowCredits(showCredits);
     
-    String curriculumName = guiderController.getCurriculumName(user.getCurriculumIdentifier());
+    return Response.ok(torWorkspaceWithCourseCredits).build();
+  }
+  
+  private String getCurriculumName(SchoolDataIdentifier curriculumIdentifier){
+    Map<SchoolDataIdentifier, String> curriculumNames = new HashMap<>();
```
aviljakaine, line 714: This will always create a new **empty** Map for every call of `getCurriculumName` kind of defeating the purpose why we want to add them to a Map in the first place. 

Using a separate function for it is fine approach (avoids repeating the if-block everywhere the name is needed), but the Map has to come in as a parameter from the calling function i.e. `private String getCurriculumName(Map<SchoolDataIdentifier, String> curriculumNameCache, SchoolDataIdentifier curriculumIdentifier) {`. After all, we want that if in the loop there are two or more calls with the same curriculumIdentifier, we use the value from the Map instead of loading it through guiderController.

You can verify it works this way with debug before changing this.

aviljakaine (dismissed):
One more change to the "cache"-system, otherwise looking ok to me

pkukkon (approved):

pkukkon (changes requested):
Some backend changes.

pkukkon (approved):
Backend OK.

PR #6875, opened 2023-11-01
Tab component refactored to be more accessible

TeroOtavia:
Refactored tabs component to be more accessible:
- swiper has now localized aria labels
- controlling swiper relies on instance
- swiper pagination pulled up from slides to be more accessible
- desktop tabs have aria labels and keyboard support
- some minor changes to other component when using tabs. Headers/titles in studies view removed because tabs component already shows those

Resolves: #6851 #6859


avukkonen (changes requested):

jjlankinen (changes requested):
After all the tests are fixed there still is serious problem regarding html 5 spec violation.

Previously we rendered conditionally Swiper or tabs component based on the breakpoint value. This change removes the conditional rendering and when doing so creates possibility of invalid html when both Swiper and tab panel's contents will be added to the document tree at the same time. This should not happen as those tab panels could contain ie. form elements which have unique id's which would not be unique anymore as both, Swiper and tabs panels would contain them at the same time.

EDIT: There's a potential fix that should work to prevent Swiper's and tabs coexistent.

jjlankinen (approved):

PR #6906, opened 2023-11-20
* Added role="button" to button component

### LINKED ISSUE #6905: Set correct role for button component's elements
We have button component that is not an actual button element but instead gets rendered as anchor. We need to add role="button" to every instance of that component's outputted elements so screen reader and keyboard users can use any of those buttons.

jjlankinen:
Closes #6905 #6859

Tabs are now buttons and there fore they can be used via keyboard only. Tabs and tab panels are linked together via aria attributes (aria-controls and aria-labelledby) and active tab is indicated as selected via aria-selected="true" attribute.

Also notebook and ToC icons have correct role and aria-label's have been added.

Added also couple of missing locale strings regarding PDF exporting.


jjlankinen:
Closing this PR as relevant and related changes are done in different PR.

PR #6921, opened 2023-12-13
6916 kuvan lisääminen sivulle ei onnistu tietyssä tilanteessa

TeroOtavia:
Changes:
- Material attachments and images can be now added even after material title name has been changed
- Attachment url paths are updated as well when material title name has been changed
- Attachment urls are now hidden

Resolves: #6916 #5002


pkukkon (approved):
Seems to work functionally.

jjlankinen (approved):
Manually tested and works just as expected.

PR #6930, opened 2024-01-17
Cleanup of unused properties in User model

aviljakaine:
Removed unused properties and an unused endpoint.

Check that these are truly unused - they seem so to me.

### DIFF
--- a/muikku-core-plugins/src/main/resources/META-INF/resources/swagger.yaml
+++ b/muikku-core-plugins/src/main/resources/META-INF/resources/swagger.yaml
@@ -14899,20 +14899,12 @@ components:
           type: string
         hasImage:
           type: boolean
-        hasEvaluationFees:
-          type: boolean
         curriculumIdentifier:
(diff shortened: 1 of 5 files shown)


muikku-core-plugins/src/main/resources/META-INF/resources/swagger.yaml
```diff
@@ -14923,23 +14915,9 @@ components:
...
           type: string
-        language:
-          type: string
-        municipality:
-          type: string
-        nationality:
-          type: string
-        school:
-          type: string
-        studyStartDate:
```
pkukkon, line 14934: Why are studyStartDate and studyTimeEnd removed from the Swagger model as well?

aviljakaine:
Follow-up issue https://github.com/otavanopisto/muikku/issues/6931

TeroOtavia (approved):

pkukkon (approved):

PR #6933, opened 2024-01-22
React version update testing

TeroOtavia:
React version updated to 17.0.2 because its peerDependency for possible comming up text engine which will with slate.js replace ckeditor. There is some errors, but for some reasons these don't cause application to fail or any issues or prevent application working anyways.

Resolves: #6985


TeroOtavia:
React Router gives error `...Render methods should be a pure function of props and state.` as container components that are charged to handle routing and passing renderMethods each route. These renderMethods do include side-effects and thats why this error is given. Side-effects should be handled in component itself (lifecycle methods) or some other way (custom hook for example). In this version of React and newer, React has been become more specific with error messages how things should be done. It doesn't break anything at least currently.

Other than this, everything seems to work fine.

avukkonen (approved):

PR #6963, opened 2024-02-14
6903 pedagogy workspaces

TeroOtavia:
Pedagogy form workspace endpoint replaced with Pedagogy form specific endpoint only returns only simple info of workspaces that student has completed or has on ongoing.

Resolves: #6903


aviljakaine (commented):
Backend ok

jjlankinen (approved):

avukkonen (approved):

PR #6981, opened 2024-02-27
6932 kurssin tuottajien nimet vaihtelevat listassa paikkaa

TeroOtavia:
Workspace producers list endpoint rework:
- Endpoint now accepts producers as list of string and returns newly created producers list back
- Simplifies frontend and backend code
- Delete endpoint removed as its not anymore needed
- Producers order in list won't be affected how fast or slow backend handles multiple single endpoint calls

Resolves: #6932


aviljakaine (approved):
Backend ok

jjlankinen (approved):

Issue #7048, opened 2024-04-24
Dialog FocusTrap V2


TeroOtavia:
Add focus trap functionality and close order functionalities if multiple dialogs are open at the same time.

child of https://github.com/otavanopisto/muikku/issues/6835

PR #7049, opened 2024-04-24
7048-changes:

TeroOtavia:
- focus trap configuration changed, should now work as expected with multiple dialogs etc...
- changes affects to dialog, slide-drawer and environment-dialog components
- portal component if closeOnEscape is true, will now close portal in order of creation

Resolves: #7048


avukkonen (approved):

jjlankinen (approved):
Testattu lokaalissa ja hyvin toimii.

PR #7261, opened 2024-12-17
7163 hops taustatiedot

TeroOtavia:
Background and Post graduate and career plan added to HOPS:
- two new tab views in the HOPS (guider including), both using wizard
- added notifications for users  that have reserved HOPS for editing purposes
- websocket support for hops form and hops form history
- react router updated to newer version.

Resolves: #7163


saaralepisto (approved):
Backend ok

avukkonen (changes requested):

avukkonen (approved):

PR #7555, opened 2025-12-17
Random bug fixes and improvments

TeroOtavia:
Random smaller bug and other improvments.

Changes:
- Dist folder is now pre-emptied before new files are placed
- Chat bubble is hidden when print styles are used
- Guardian records side also include student's credits same way as in records view
- Prevented Help materials from crashing if pasted material was some sort of assignment type. Note that there is still no point to paste these kind of materials to help anyways.

Resolves: #7358 #7374 #7126


muikku-frontend/muikkuV3/package.json
```diff
@@ -87,6 +87,7 @@
   },
   "devDependencies": {
     "@hot-loader/react-dom": "17.0.2",
+    "@types/fs-extra": "^11.0.4",
```
jjlankinen: Hattu pois versionumerosta.

PR #7725, opened 2026-08-27
7719 refactor memo field to use only rich content editor

### LINKED ISSUE #7719: Refactor memo field to use only rich-content editor
Refactor memo field to use only rich-content editor

TeroOtavia:
Memofield is now always considered as rich content editor field, so it uses Ckeditor for now on. Existing memofields that are defined to use textarea and already have existing data are automatically converted to rich content with as best possible guess logic available.

Memofield character and word limits are now handled through cke own events and event is cancelled if limit is hit.

Also, weird ckeditor crash with memo and journal fields that was caused by its own unpredictable compatible with React logic is now fixed.

This branch should be merged after #7697 is in produciton and is confirmed that it works fine

Resolves: #7719


muikku-frontend/muikkuV3/components/general/ckeditor.tsx
```diff
@@ -52,37 +53,64 @@ const PLUGINS = {
...
-   */
+  // eslint-disable-next-line jsdoc/require-jsdoc
   cancel(): void;
-  /**
-   * stop method
-   */
-  stop(): void;
 }
 
+/**
```
avukkonen: commented code

avukkonen (changes requested):

avukkonen (approved):

